PR #1, opened 2016-02-16
First pull request

seanh:
Working proof of concept, given suitable changes to your local Chrome extension (and hacking this repo's code to include your extension's ID instead of mine), this will redirect you to the extension or (if you disable the extension) to Via.

- [x] Read config settings from env vars (Chrome extension ID, Elasticsearch URL...)
- [x] Add linter configs
- [x] Write Python tests
- [x] Write JavaScript tests
- [x] Add Travis CI
- [ ] Add Coveralls
- [ ] Add Landscape
- [x] Add Hound
- [x] Write a CONTRIBUTING.rst
- [x] Add the CSS redirecting animation
- [x] Get a better Hypothesis icon for the redirecting page (SVG would be nice)
- [x] Add a jinja scripts block in the right place
- [ ] Handle query params, anchors in document URI
- [x] Run in Python 3 not 2.7
- [ ] Test it on a wide range of annotations, make sure that the code for extracting the annotated URI always works
- [x] Alphabetize CSS
- [x] Docker support for production deployment
- [x] <del>Include font</del>
- [ ] Issue with image not showing (if JavaScript loads and executes before image loads?)
- [x] Fix browserify / JavaScript assets issue

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,6 @@
+javascript:
+  config_file: .jshintrc
+  ignore_file: .jshintignore
+jscs:
+  enabled: true
+  config_file: .jscsrc
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,30 @@
+FROM frolvlad/alpine-python3
+MAINTAINER Hypothes.is Project and contributors
+
+# Install system build and runtime dependencies.
+RUN apk add --update \
+    nodejs \
+  && rm -rf /var/cache/apk/*
+
+# Create the bouncer user, group, home directory and package directory.
+RUN addgroup -S bouncer \
+  && adduser -S -G bouncer -h /var/lib/bouncer bouncer
+WORKDIR /var/lib/bouncer
+
+# Copy packaging
+COPY README.rst package.json setup.* ./
+
+# Install application dependencies.
+RUN npm install --production \
+  && pip install --no-cache-dir -e . \
+  && npm cache clean
+
+# Copy the rest of the application files
+COPY bouncer ./bouncer/
+
+# Persist the static directory.
+VOLUME ["/var/lib/bouncer/bouncer/static"]
+
+# Start the web server by default
+USER bouncer
+CMD ["gunicorn", "bouncer:app()"]
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+Copyright (c) 2016 Hypothes.is Project and contributors
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/bouncer/__init__.py
+++ b/bouncer/__init__.py
@@ -1,0 +1,43 @@
+import os
+
+import pyramid.config
+
+
+def settings():
+    """
+    Return the app's configuration settings as a dict.
+
+    Settings are read from environment variables and fall back to hardcoded
+    defaults if those variables aren't defined.
+
+    """
+    via_base_url = os.environ.get("VIA_BASE_URL", "https://via.hypothes.is")
+    if via_base_url.endswith("/"):
+        via_base_url = via_base_url[:-1]
+
+    return {
+        "chrome_extension_id": os.environ.get(
+            "CHROME_EXTENSION_ID",
+            "bjfhmglciegochdpefhhlphglcehbmek"),
+        "elasticsearch_host": os.environ.get("ELASTICSEARCH_HOST",
+                                             "localhost"),
+        "elasticsearch_index": os.environ.get("ELASTICSEARCH_INDEX",
+                                              "annotator"),
+        "elasticsearch_port": os.environ.get("ELASTICSEARCH_PORT",
+                                             "9200"),
+        "hypothesis_url": os.environ.get("HYPOTHESIS_URL",
+                                         "https://hypothes.is"),
+        "via_base_url": via_base_url,
+    }
+
+
+def app():
+    """Configure and return the WSGI app."""
+    config = pyramid.config.Configurator(settings=settings())
+    config.add_static_view(name="static", path="static")
+    config.include("pyramid_jinja2")
+    config.registry.settings["jinja2.filters"] = {
+        "static_url": "pyramid_jinja2.filters:static_url_filter"
+    }
+    config.include("bouncer.views")
+    return config.make_wsgi_app()
--- a/bouncer/templates/base.html.jinja2
+++ b/bouncer/templates/base.html.jinja2
@@ -1,0 +1,16 @@
+<!doctype html>
+  <html lang=en>
+  <head>
+    <meta charset=utf-8>
+    <title>{{ title|safe }}</title>
+    <link rel="stylesheet"
+          type="text/css"
+          href="{{'bouncer:static/styles/bouncer.css'|static_url}}">
+  </head>
+  <body>
+    <div class="center">
+      {% block content %}{% endblock %}
+    </div>
+  </body>
+  {% block scripts %}{% endblock %}
+</html>
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -1,0 +1,78 @@
+import json
+
+import elasticsearch
+from elasticsearch import exceptions
+from pyramid import httpexceptions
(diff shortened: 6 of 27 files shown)


.hound.yml
```diff
@@ -0,0 +1,9 @@
+coffeescript:
```
nickstenning: I'm pretty sure if there are no `.coffee` files in this repo this isn't needed.

.hound.yml
```diff
@@ -0,0 +1,9 @@
...
+  enabled: false
+javascript:
+  config_file: .jshintrc
+  ignore_file: .jshintignore
+ruby:
```
nickstenning: Ditto.

.hound.yml
```diff
@@ -0,0 +1,9 @@
...
+  enabled: false
+scss:
```
nickstenning: Ditto.

LICENSE
```diff
@@ -0,0 +1,21 @@
+Copyright (c) 2013-2014 Hypothes.is Project and contributors
```
nickstenning: Might want to update the dates here.

bouncer/__init__.py
```diff
@@ -0,0 +1,41 @@
+import os
+
+import pyramid.config
+
+
+def settings():
```
nickstenning: It's a small thing, but I would tend to avoid naming getters like this. If I saw a `settings` object in this module, I probably wouldn't assume it was callable. If I saw a `get_settings` object in the module, I definitely *would*.

bouncer/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+        "elasticsearch_index": os.environ.get("ELASTICSEARCH_INDEX",
+                                              "annotator"),
+        "chrome_extension_id": os.environ.get(
+            "CHROME_EXTENSION_ID",
+            "bjfhmglciegochdpefhhlphglcehbmek"),
+        "via_base_url": via_base_url,
+    }
+
+
+def app():
```
nickstenning: Similarly, I'd suggest `create_app` is clearer.

bouncer/static/scripts/redirect.js
```diff
@@ -0,0 +1,24 @@
...
+
+function getSettings(document) {
+  return JSON.parse(
+    document.querySelector('script.js-bouncer-settings').textContent);
+}
+
+// The ID of the extension we want to talk to.
+var editorExtensionId = 'oldbkmekfdjiffgkconlamcngmkioffd';
+
+chrome.runtime.sendMessage(editorExtensionId, 'Hi?', function (response) {
```
nickstenning: As per my comment on the other repo, `PING` would be more conventional :)

nickstenning: You might want to check that `chrome.runtime.sendMessage` exists before you try and call it!

bouncer/templates/base.html.jinja2
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+  <html lang=en>
+  <head>
+    <meta charset=utf-8>
+    <title>{% trans %}Hypothesis Direct-Link Bouncer Service{% endtrans %}</title>
```
nickstenning: I'd suggest replacing this with a more user-friendly string. Perhaps `Hypothesis: redirecting you...`?

bouncer/views.py
```diff
@@ -0,0 +1,77 @@
...
+def elasticsearch_client(settings):
+    return elasticsearch.Elasticsearch(
+        host=settings["elasticsearch_host"],
+        port=settings["elasticsearch_port"],
+    )
+
+
+@view.view_config(renderer="bouncer:templates/index.html.jinja2",
+                  route_name="index")
+def index(request):
```
nickstenning: This should just redirect to `https://hypothes.is`.

setup.py
```diff
@@ -0,0 +1,66 @@
...
+        "pyramid==1.6.1",
+        "pyramid-jinja2==2.6.2",
+        "elasticsearch>=2.0.0,<3.0.0",
+    ],
+
+    extras_require={
+        'dev': DEV_EXTRAS,
+    },
+
+    package_data={
```
nickstenning: Could we just omit these rather than having them empty here?

setup.py
```diff
@@ -0,0 +1,66 @@
...
+    keywords="annotation web javascript",
+    license="Simplified (2-Clause) BSD License",
+    long_description=long_description,
+    name="bouncer",
+    url="https://github.com/hypothesis/bouncer",
+    version="0.0.1",  # PEP440-compliant version number.
+
+    # See https://pypi.python.org/pypi?%3Aaction=list_classifiers
+    classifiers=[
+        # How mature is this project? Common values are
```
nickstenning: I don't think this comment adds much. It says "Alpha" right in the classifier.

setup.py
```diff
@@ -0,0 +1,66 @@
+from setuptools import setup, find_packages
+from codecs import open
```
nickstenning: This isn't needed in Python 3 I don't think. You can just use the `open` builtin, as the default character encoding in Python 3 is UTF-8.

nickstenning:
Looking pretty good. I would only request that you rebase out the import/removal of stuff from `h` as it looks a bit messy in the history.

seanh:
Alright alright! This isn't ready for review! I wouldn't submit it for review in its current state. But since I'm working a Saturday, the comments will give me plenty to chew on

nickstenning:
Oh, sorry. I figured if you had opened a PR you were looking for a review.

bouncer/templates/base.html.jinja2
seanh: Sure

setup.py
seanh: Done

setup.py
seanh: Done

LICENSE
seanh: Done

seanh:
Ok, this is ready for review now.

As you can see from the task list at the top I'd still like to get Landsape and Coveralls enabled. Also there is Python code in here to extract the annotated document's URI from an Elasticsearch annotation document. I haven't tested that on a wide variety of production annotation data. Do all annotation documents have the top-level "uri" field? Do some of the URIs contain things like query tags or fragments that we might want to strip?

I'll copy-paste th

.hound.yml
seanh: This was copy-pasted from h and I hadn't edited it yet, it's gone now.

.hound.yml
seanh: Ditto.

.hound.yml
seanh: Ditto.

bouncer/__init__.py
seanh: This is only ever used once though, and like this: `settings=settings()`. Is there really potential for confusion? I kind of like the simpler name.

bouncer/__init__.py
seanh: Only used by `Makefile` and looks like this: `gunicorn --reload "bouncer:app()"`

bouncer/static/scripts/redirect.js
seanh: > You might want to check that chrome.runtime.sendMessage exists before you try and call it!

Done, non-Chrome browsers should now redirect to Via without error.

bouncer/static/scripts/redirect.js
```diff
@@ -0,0 +1,58 @@
...
+  navigateTo = navigateTo_ || navigateTo;
+
+  var settings = getSettings(document);
+
+  if (window.chrome) {
+    // The user is using Chrome, redirect them to our Chrome extension if they
+    // have it installed, via otherwise.
+    chrome.runtime.sendMessage(
+      EDITOR_EXTENSION_ID,
+      'Hi?',
```
robertknight: I suggest it might be a good idea to exchange structured objects with the extension so that you have room to add additional information if you need in future. The pattern I've been following is that all messages sent back and forth between scripts are objects with a 'type' property which is a string identifying the message, plus additional info as needed.

bouncer/static/scripts/redirect.js
```diff
@@ -0,0 +1,58 @@
...
+
+/** Navigate the browser to the requested annotation.
+ *
+ * If the browser is Chrome and our Chrome extension is installed then
+ * navigate to the annotation's direct link for the Chrome extension.
+ * If the Chrome extension isn't installed or the browser isn't Chrome then
+ * navigate to the annotation's Via direct link.
+ *
+ */
+function redirect(navigateTo_) {
```
robertknight: The convention we have for naming arguments that are functions is a `Fn` suffix, so `navigateFn`.

bouncer/static/scripts/redirect.js
```diff
@@ -0,0 +1,58 @@
...
+  // Use the test navigateTo() function if one was passed in, the real
+  // navigateTo() otherwise.
+  navigateTo = navigateTo_ || navigateTo;
+
+  var settings = getSettings(document);
+
+  if (window.chrome) {
```
robertknight: You need to test for the existence of `chrome.runtime` and `chrome.runtime.sendMessage` explicitly. Microsoft Edge for example has a `window.chrome` object with only a handful of properties.

seanh:
I've fixed the Docker Python 3 issue by using someone's Alpine Linux Python 3 image.

And fixed the browserify issue by just running browserify in the tests only, as suggested by Rob. This means that redirect.js is inlined into the HTML too (which is just done by Jinja).

bouncer/static/scripts/redirect.js
seanh: Done

bouncer/static/scripts/redirect.js
seanh: Done

bouncer/static/scripts/redirect.js
seanh: Done

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM frolvlad/alpine-python3
```
nickstenning: Is there a reason you've used this rather than just installing python3 using apk?

seanh: I did try to get an `apk add python3` version working, but it would still end up running `setup.py` with Python 2 and crashing.

seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;wLglsAHZ&#x2F;258-build-standalone-service-which-can-dispatch-to-extension-or-via-as-appropriate"> Build standalone service which can dispatch to extension or via as appropriate.</a>

nickstenning:
LGTM. I think we may want to revisit the Dockerfile -- not too keen on using non-official images -- but other than that great work.

First Python 3 in our infrastructure! :tada:

Dockerfile
chdorner: @seanh I guess the reason why it still used Python 2 is because the python-pip package depends on Python 2. I can have a look at that so that we use the same base image as with H if that's still a concern?

nickstenning: https://github.com/hypothesis/smokey/blob/master/smokey/Dockerfile#L4

seanh: >  guess the reason why it still used Python 2 is because the python-pip package depends on Python 2.

That seemed to be what was happening, yes. I'd apk install Python 3 and _not_ Python 2, but then my `setup.py` would get run in Python 2 and crash, presumably because it was Python 2 pip that got installed.

Looks like nick has the solution, I'll open an issue so that I remember to fix this

seanh: https://github.com/hypothesis/bouncer/issues/14

I'll do this at some point

PR #5, opened 2016-02-29
Report some initial stats to statsd

seanh:
Statsd is configured using environment variables documented in the README.

Each view callable reports its result to statsd using a naming convention based on [this advice](https://matt.aimonetti.net/posts/2013/06/26/practical-guide-to-graphite-monitoring/), e.g. "views.annotation.404.annotation_not_found", "views.annotation.200".

Separately a `NewRequest` subscriber increments a "requests" counter that simply counts the number of incoming requests regardless of which view (if any) gets called and what its outcome is.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,4 +27,4 @@ VOLUME ["/var/lib/bouncer/bouncer/static"]
 
 # Start the web server by default
 USER bouncer
-CMD ["gunicorn", "bouncer:app()"]
+CMD ["gunicorn", "--config", "gunicorn.conf.py", "bouncer:app()"]
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ deps:
 	npm install
 
 dev:
-	PYRAMID_RELOAD_TEMPLATES=1 gunicorn --reload "bouncer:app()"
+	PYRAMID_RELOAD_TEMPLATES=1 gunicorn --config gunicorn.conf.py --reload "bouncer:app()"
 
 docker:
 	docker build -t hypothesis/bouncer .
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -4,6 +4,7 @@ import elasticsearch
 from elasticsearch import exceptions
 from pyramid import httpexceptions
 from pyramid import view
+from statsd.defaults.env import statsd
 
 
 @view.view_defaults(renderer="bouncer:templates/annotation.html.jinja2")
@@ -15,7 +16,6 @@ class AnnotationController(object):
     @view.view_config(route_name="annotation_with_url")
     @view.view_config(route_name="annotation_without_url")
     def annotation(self):
-
         settings = self.request.registry.settings
 
         try:
@@ -24,6 +24,7 @@ class AnnotationController(object):
                 doc_type="annotation",
                 id=self.request.matchdict["id"])
         except exceptions.NotFoundError:
+            statsd.incr("views.annotation.404.annotation_not_found")
             raise httpexceptions.HTTPNotFound()
 
         annotation_id = document["_id"]
@@ -39,6 +40,7 @@ class AnnotationController(object):
         extension_url = "{uri}#annotations:{id}".format(
             uri=document_uri, id=annotation_id)
 
+        statsd.incr("views.annotation.200.annotation_found")
         return {
             "data": json.dumps({
                 # Warning: variable names change from python_style to
@@ -59,6 +61,7 @@ def elasticsearch_client(settings):
 @view.view_config(renderer="bouncer:templates/index.html.jinja2",
                   route_name="index")
 def index(request):
+    statsd.incr("views.index.302.redirected_to_hypothesis")
     raise httpexceptions.HTTPFound(
         location=request.registry.settings["hypothesis_url"])
 
--- a/README.rst
+++ b/README.rst
@@ -12,6 +12,55 @@
 Hypothesis Direct-Link Bouncer Service
 ======================================
 
+Configuration
+-------------
+
+You can set various environment variables to configure bouncer before running
+it in production or development:
+
+CHROME_EXTENSION_ID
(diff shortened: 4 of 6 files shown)


bouncer/views.py
```diff
@@ -67,6 +72,7 @@ def index(request):
     append_slash=True,
     renderer='bouncer:templates/notfound.html.jinja2')
 def notfound(request):
+    statsd.incr("views.404.path_did_not_match_any_view_config")
```
nickstenning: I wouldn't bother logging the generic 404, as we'll get that from `gunicorn` stats anyway.

bouncer/views.py
```diff
@@ -24,6 +26,7 @@ def annotation(self):
                 doc_type="annotation",
                 id=self.request.matchdict["id"])
         except exceptions.NotFoundError:
+            statsd.incr("views.annotation.404.annotation_not_found")
```
nickstenning: I'd suggest not bothering with this. 404s will be counted by gunicorn.

seanh:
Coverage is down mostly because of the introduction of the `subscribers` module. I could add a unit test for this but it doesn't seem necessary.

bouncer/views.py
```diff
@@ -39,6 +42,7 @@ def annotation(self):
         extension_url = "{uri}#annotations:{id}".format(
             uri=document_uri, id=annotation_id)
 
+        statsd.incr("views.annotation.200")
```
nickstenning: I'd be inclined to call this `views.annotation` on its own. We'll get stats on statuses from gunicorn.

bouncer/views.py
```diff
@@ -59,6 +63,7 @@ def elasticsearch_client(settings):
 @view.view_config(renderer="bouncer:templates/index.html.jinja2",
                   route_name="index")
 def index(request):
+    statsd.incr("views.index.302")
```
nickstenning: Ditto. Just `views.index`.

bouncer/subscribers.py
```diff
@@ -0,0 +1,12 @@
...
+from pyramid import events
+from statsd.defaults.env import statsd
+
+
+def increment_requests_counter(event):
+    statsd.incr("requests")
+
+
+def includeme(config):
+    config.add_subscriber(increment_requests_counter, events.NewRequest)
```
nickstenning: Also not needed as we'll get this from gunicorn.

bouncer/views.py
seanh: Will gunicorn tell us the difference between a generic 404 vs an annotation not found?

nickstenning: No. I suppose having the annotation not found as a separate stat is useful, but perhaps put it in its own hierarchy rather than tying it to a view. Or, if you do want them to be attached to the views then make it easy to aggregate over all the requests a view gets, which you could do by ensuring that all view events are of the form `views.<name>.<status>`. Having an extra part of the hierarchy (like `.annotation_not_found` in this case) makes it tricky to aggregate.

seanh:
Ok, looking at [gunicorn's instrumentation](http://docs.gunicorn.org/en/latest/instrumentation.html) it reports:

* `gunicorn.requests`: global request rate (what I did with the `NewRequest` subscriber)
* `gunicorn.request.status.{200,302,304,404,...}`: I'm not sure how useful this is though, it only counts the number of 404s (for example) not which view 404'd or why
* `gunicorn.request.duration`: response time, looks like we won't need to measure that ourselves.

It looks to me like it still might be useful for our Python code to report to statsd:

* The number of times each view callable is hit, I think this does include the notfound view because that's counting 404's due to no view config matching the path, which are different from 404s due the annotation view not finding an annotation matching the requested ID. Gunicorn tells us how many requests are coming in to bouncer, this additionally tells us how many of those are:

   * For the index page
   * A 404 that didn't match any view config
   * Requests that actually match the annotation route

  This doesn't seem _super_ useful but it does give us a bit more over just gunicorn.

* Within each view callable, what its result is (for example 200 found an annotation vs 404 did not find an annotation with that ID). The annotation view is the only view that has multiple outcomes at the moment. So this will tell us:

   * How many requests actually get redirected to an annotation (or at least, how many receive a response with HTML and JavaScript that will try to redirect them). This is the closest thing the server can give us to the number of actual successful redirects.

   * How many requests match the annotation route but get 404'd because bouncer couldn't find an annotation matching the requested ID. This would happen if someone shared a direct link to an annotation, then deleted the annotation, but people were still clicking on the link. Or if someone has a typo or copy-paste mistake with a link. Or if some person or bot is constructing invalid direct link URLs. I'd expect this to be a low number but it might tell us something.

seanh:
@nickstenning This can be reviewed again

Dockerfile
```diff
@@ -27,4 +27,4 @@ VOLUME ["/var/lib/bouncer/bouncer/static"]
 
 # Start the web server by default
 USER bouncer
-CMD ["gunicorn", "bouncer:app()"]
+CMD ["gunicorn", "--config", "gunicorn.conf.py", "bouncer:app()"]
```
nickstenning: This isn't needed (the default is to load a `gunicorn.conf.py` in CWD).

nickstenning: Oh, but if you want to use it, you probably want to `COPY` it into the image.

Makefile
```diff
@@ -5,7 +5,7 @@ deps:
 	npm install
 
 dev:
-	PYRAMID_RELOAD_TEMPLATES=1 gunicorn --reload "bouncer:app()"
+	PYRAMID_RELOAD_TEMPLATES=1 gunicorn --config gunicorn.conf.py --reload "bouncer:app()"
```
nickstenning: Again, not needed.

nickstenning:
LGTM.  A couple of comments you might want to follow up on.

Dockerfile
seanh: Hmm, we will definitely have to do that

PR #6, opened 2016-02-29
Report Python exceptions to Sentry

seanh:
Report Python exceptions to a Sentry project.

- Raven (the official Python client for Sentry) is added to the
  requirements in setup.py.

- The Sentry DSN (the URI that's needed to authenticate to Sentry) is
  read from the SENTRY_DSN environment variable like all other config
  settings in bouncer.

- bouncer/sentry.py is a new module that contains all the new
  Sentry/raven stuff.

  We add a raven client object configured with the SENTRY_DSN from the
  environment as request.raven, so that it can be accessed by our raven
  tween.

  Then we wrap the existing bouncer Pyramid app in a Pyramid tween that
  catches any exceptions, reports them to Sentry, then re-raises the
  exception.

  Any uncaught exceptions raised anywhere in the code, e.g. in a view
  callable, will be caught and reported to Sentry by this tween.

### DIFF
--- a/bouncer/__init__.py
+++ b/bouncer/__init__.py
@@ -40,4 +40,5 @@ def app():
         "static_url": "pyramid_jinja2.filters:static_url_filter"
     }
     config.include("bouncer.views")
+    config.include("bouncer.sentry")
     return config.make_wsgi_app()
--- a/bouncer/sentry.py
+++ b/bouncer/sentry.py
@@ -1,0 +1,53 @@
+"""Report exceptions to Sentry/Raven."""
+from pyramid import tweens
+import raven
+
+from bouncer import __about__
+
+
+#: The Raven client object that's used to send reports to Sentry.
+#: This object is created once per app and re-used across requests.
+#:
+#: This should be accessed via request.raven rather than directly, because
+#: get_raven_client() below adds per-request context to the client.
+raven_client = raven.Client(release=__about__.__version__)
+
+
+def raven_tween_factory(handler, _):
+    """Return a tween that reports uncaught exceptions to Sentry."""
+    def raven_tween(request):
+        """Report uncaught exceptions to Sentry."""
+        try:
+            return handler(request)
+        except:
+            request.raven.captureException()
+            raise
+
+    return raven_tween
+
+
+def get_raven_client(request):
+    """Add per-request context to raven_client and return it."""
+    # Add request-specific context to the Raven client.
+    raven_client.http_context({
+        "url": request.url,
+        "method": request.method,
+    })
+
+    # Clear the request-specific context at the end of each request so that it
+    # isn't carried over to the next request (the same raven_client object is
+    # used across requests).
+    request.add_finished_callback(
+        lambda request: raven_client.context.clear())
+
+    return raven_client
+
+
+def includeme(config):
+    config.add_request_method(
+        get_raven_client,
+        name="raven",
+        reify=True)
+    config.add_tween(
+        "bouncer.sentry.raven_tween_factory",
+        under=tweens.EXCVIEW)
--- a/README.rst
+++ b/README.rst
@@ -38,6 +38,10 @@ HYPOTHESIS_URL
   The URL of the Hypothesis front page that requests to bouncer's front page
   will be redirected to (default: https://hypothes.is)
 
+SENTRY_DSN
+    The DSN (Data Source Name) that bouncer will use to report crashes to
+    `Sentry <https://getsentry.com/>`_
+
 STATSD_HOST
   The host of the statsd server that bouncer will report stats to
   (default: localhost)
--- a/bouncer/__about__.py
+++ b/bouncer/__about__.py
@@ -1,0 +1,6 @@
+"""Metadata about bouncer shared between setup.py and bouncer code."""
+
+__all__ = ["__version__"]
+
+
+__version__ = "0.0.1"  # PEP440-compliant version number.
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -67,7 +67,6 @@ def index(request):
 
 
 @view.notfound_view_config(
-    append_slash=True,
     renderer='bouncer:templates/notfound.html.jinja2')
 def notfound(request):
     request.response.status_int = 404
@@ -77,5 +76,5 @@ def notfound(request):
 def includeme(config):
     config.add_route("index", "/")
     config.add_route("annotation_with_url", "/{id}/*url")
-    config.add_route("annotation_without_url", "/{id}/")
+    config.add_route("annotation_without_url", "/{id}")
     config.scan(__name__)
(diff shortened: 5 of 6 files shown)


seanh:
I should add `SENTRY_DSN` to the README but I'm waiting on some README refactorings in https://github.com/hypothesis/bouncer/pull/5 to be merged first

seanh:
I'd also like to go through the docs and add any useful Sentry options that I'm not using.

JavaScript reporting to Sentry to come in a separate pull request.

bouncer/sentry.py
```diff
@@ -0,0 +1,31 @@
...
+        except:
+            request.raven.captureException()
+            raise
+
+    return raven_tween
+
+
+def raven_client(request):
+    """Return the raven.Client object."""
+    request.add_finished_callback(
```
nickstenning: Either you create the client once and you need this callback, or you create the client every request and you don't. I'd suggest the former.

bouncer/__init__.py
```diff
@@ -27,6 +27,7 @@ def settings():
                                              "9200"),
         "hypothesis_url": os.environ.get("HYPOTHESIS_URL",
                                          "https://hypothes.is"),
+        "sentry_dsn": os.environ.get("SENTRY_DSN"),
```
nickstenning: This isn't needed AIUI. Raven is already capable of reading the setting from the environment.

bouncer/sentry.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+
+    return raven.Client(
+        dsn=request.registry.settings.get("sentry_dsn"),
+    )
+
+
+def includeme(config):
+    config.add_request_method(raven_client, name="raven", reify=True)
+    config.add_tween("bouncer.sentry.raven_tween_factory")
```
nickstenning: A call to `add_tween` without specifying `under` or `over` hints means:

> Specifying neither over nor under is equivalent to specifying under=INGRESS.

(From [the Pyramid docs](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html))

That means that the ordering of tweens here will be:

```
INGRESS
raven_tween_factory
EXCVIEW
MAIN
```

Which in turn means that the pyramid excview will catch all the exceptions before they get to Raven. You probably want to specify `under=EXCVIEW` here.

nickstenning: Ah, although I note that we actually don't have any catch-all exception views in this application. So actually this is probably fine for now.

nickstenning:
So two (and a half) things here:

- [ ] No need to pass the setting through `config.registry.settings`. Just let `raven.Client` read it from the environment.
- [ ] Create the client once per application rather than once per request.
- [ ] Rebase.

seanh:
I've [added basic info about the request](https://github.com/hypothesis/bouncer/commit/63879bea7ffc6c4f2058a7d19a1c09fa17ed6a47) and [release version number](https://github.com/hypothesis/bouncer/commit/2409e8c77657d2851aaf988c0a91a8679cdba9d0) to the Sentry reports.

Any other info worth adding to them? We don't use anything from the request other than the path, we don't have user accounts (and I didn't think IP addresses in Sentry would be useful), and I can't think of any tags or extra data that would be useful right now, so I think that's enough.

bouncer/__init__.py
seanh: So it is

bouncer/sentry.py
seanh: Yeah, it _is_ working (I'm testing against my own hosted Sentry account) but I may as well add this just so we don't break it if we add an exception view

nickstenning: Well, it's tricky. I'd leave it as is. The problem is that if your exception view is a catch-all that swallows exceptions, then it will break raven. But if it's catching a specific kind of exception which doesn't need reporting, then raven should be over it, not under it.

bouncer/sentry.py
seanh: I need the request to create the client. The request was being used to access the Pyramid config registry to pass settings to the client. That's no longer necessary, but I'm now using the request to add context (requested URL etc) to the client. I _could_ have a `NewRequest` subscriber add that context instead of doing it in this `raven_client()` function, then I could create the client once at module scope and use the `context.clear()` finished callback. But that spreads code about initialising and configuring the raven client across three different places. Worth the trouble?

I think maybe just creating and configuring new client for each request (and getting rid of the context.clear() as it's unnecessary) is simpler, and expect it doesn't add much workload.

bouncer/sentry.py
seanh: Didn't see this comment. I've [tweaked things so that 404s caught by append_slash=True are no longer reported to Sentry](https://github.com/hypothesis/bouncer/commit/fe76bc6204bc0d76d583cf9fe0c788bd5792bd24). 
If we add an exception view that swallows certain kinds of exceptions that don't need reporting - those would still get reported to Sentry.

However if we just wanted to add an exception view to make it more pretty or useful when something goes wrong, we could do that now and the exceptions would still be reported.

seanh:
@nickstenning All done again

bouncer/sentry.py
```diff
@@ -0,0 +1,53 @@
...
+
+from bouncer import __about__
+
+
+#: The Raven client object that's used to send reports to Sentry.
+#: This object is created once per app and re-used across requests.
+#:
+#: This should be accessed via request.raven rather than directly, because
+#: get_raven_client() below adds per-request context to the client.
+raven_client = raven.Client(release=__about__.__version__)
```
nickstenning: It would make more sense to do this when the app is build in the `includeme` and shove it on the registry, as in `warehouse`...

bouncer/sentry.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def includeme(config):
+    config.add_request_method(
+        get_raven_client,
+        name="raven",
+        reify=True)
+    config.add_tween(
+        "bouncer.sentry.raven_tween_factory",
+        under=tweens.EXCVIEW)
```
nickstenning: As you've discovered, contrary to my previous comment this actually makes life difficult. I'd suggest reverting this, and bringing back `append_slash=True`, but hey, no big deal either way.

nickstenning:
I'll leave the comments for you to address separately. Looks good! Let's get this deployed soon and iterate on it from there.

bouncer/sentry.py
seanh: I've gone ahead and [done this on master](https://github.com/hypothesis/bouncer/commit/f2fb1b6b2f76445d303753f6ba6bc7cbbcafbcb4)

bouncer/sentry.py
seanh: Either way seems about the same to me. We can either have appending slashes but if we do add an exception view one day it'll prevent crashes from being reported to Sentry. Or we can not have appending slashes but have the not-slashed URLs just work as they are, and have reporting crashes to Sentry be safe from any future exception view.

PR #7, opened 2016-03-01
Fix docker build

seanh:


seanh:
@nickstenning Deployment should now work. `make docker` works for me locally and the dockerized bouncer seems to work and report to statsd and Sentry correctly, including reading config from env vars

Dockerfile
```diff
@@ -14,17 +14,20 @@ WORKDIR /var/lib/bouncer
 # Copy packaging
 COPY README.rst package.json setup.* ./
 
+# Copy gunivorn config
+COPY gunicorn.conf.py ./
+
+# Copy the rest of the application files
+COPY bouncer ./bouncer/
```
nickstenning: The reason this was below `npm install ...` is so that you don't need to rerun the dependency installation unless the dependencies change. By putting it up here you're ensuring that the Docker image cache is busted for any change anywhere in the application. Both this and the `gunicorn.conf.py` should go below the `RUN npm install ...` line.

Dockerfile
```diff
@@ -14,17 +14,20 @@ WORKDIR /var/lib/bouncer
 # Copy packaging
 COPY README.rst package.json setup.* ./
 
+# Copy gunivorn config
```
nickstenning: Probably don't need this comment ;)

nickstenning:
Oh, i see... you've added the application code because `setup.py` now requires it. We have ways around this for `h` but we can save those for another day here.

PR #8, opened 2016-03-01
Add Sentry reporting of JavaScript crashes

seanh:
Add raven.min.js (Sentry's JavaScript client) to the annotation
redirector page (currently just using Sentry's CDN-hosted copy).

Add an additional environment variable SENTRY_JAVASCRIPT_DSN since the
Sentry DSN that the client-side JavaScript code uses is different to the
one that the server-side code uses (it doesn't contain the secret key
part because it isn't secure).

The server renders the SENTRY_JAVASCRIPT_DSN into the HTML along with
the rest of the settings.

The JavaScript code in redirect.js picks this up and uses it to
configure Raven, which automatically reports crashes to Sentry.

Also add the bouncer release version number to the settings object that
the server sends to the JavaScript so that JavaScript crash reports can
include the version number as Python ones do.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,8 +39,14 @@ HYPOTHESIS_URL
   will be redirected to (default: https://hypothes.is)
 
 SENTRY_DSN
-    The DSN (Data Source Name) that bouncer will use to report crashes to
-    `Sentry <https://getsentry.com/>`_
+    The DSN (Data Source Name) that bouncer's server-side code will use to
+    report crashes to `Sentry <https://getsentry.com/>`_
+
+SENTRY_JAVASCRIPT_DSN
+    The DSN (Data Source Name) that bouncer's client-side JavaScript code will
+    use to report crashes to `Sentry <https://getsentry.com/>`_.
+    This is different to SENTRY_DSN above - it doesn't include the secret key
+    part of the DSN.
 
 STATSD_HOST
   The host of the statsd server that bouncer will report stats to
@@ -112,7 +118,7 @@ To debug the JavaScript tests in a browser, run:
 
 .. code-block:: bash
 
-   ./node_modules/karma/bin/karma start --no-single-rin
+   ./node_modules/karma/bin/karma start --no-single-run
 
 and open http://localhost:9876/ in your browser.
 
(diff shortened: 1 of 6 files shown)


seanh:
JavaScript tests are broken by the module-level attempt to access `document` in `redirect.js`

robertknight:
Since you've got this working, I'm prepared to give it a try and see whether anything unexpected comes up. I would suggest calling it `SENTRY_PUBLIC_DSN` because that is how the documentation refers to it.

seanh:
Oh, and before I forget:

* The reporting of tracebacks in Sentry is messed up because we use Jinja2 to inline `redirect.js` into the HTML
* We might want to distribute Raven ourselves, instead of using Sentry's DSN?

nickstenning:
I'd suggest we don't do this for now, for a couple of reasons:

- Loading the raven script will block execution of `redirect.js`. `raven.min.js` is 22KiB compared with <2KiB for `redirect.js` which means that on a slow connection we'll have made the delay before executing the actual application code up to an order of magnitude greater.
- The code in `redirect.js` is small, and is only intended to do one thing, in one browser. I can't believe we're going to catch a lot of interesting stuff with Raven anyway.

seanh:
It _does_ need to work across browsers, even if in most browsers it just redirects to Via. I accept that the code is simple though.

My feeling is that early on in bouncer's deployment I'd rather have the reporting and know whether it's working, rather than have it micro-optimised for speed. My bet is that the biggest factor by far in the perceived speed of direct links is going to be the time that the page we're redirecting to takes to load, I think in most browsers our redirecting page will still be showing for a noticeable amount of time after the browser has been redirected and is downloading the next page (and even if it isn't, the time the page we redirect to takes to load is still part of the time the user has to wait until they get to the annotation). The `raven.min.js` is (currently) coming from their CDN and so is cached, for me it takes 42ms to get the 304 Not Modified response from the CDN, compared to about 8 seconds to load a small BBC news article in Via and see the sidebar appear, and about 4 seconds to load the same article directly with the Chrome extension already activated in the tab and see the sidebar appear.

Later on we can measure its performance in production and then optimise it, for example by writing our own Sentry-reporting JavaScript (if we can do it in less than 22K) and maybe inlining it into the HTML, looking into more caching (of the CSS, the SVG file), or whatever.

However, I'm perfectly happy for this to be held back if that's what others want.

nickstenning:
I suggest we hold off on this for now.

PR #10, opened 2016-03-03
Don't redirect to file:// URLs

seanh:


seanh:
I've gone with a whitelist approach here (we only redirect to http and https), because we do have some other protocols in our data besides file (`zotero://`, `view-source://`, `urn:x-pdf:`, `chrome-extension://`, ...)

We also have some documents in Elastic search whose URI is the string `"[object URL]"` and some whose uri us the value `None`. I'll need to fix bouncer not to crash on those.

bouncer/views.py
```diff
@@ -25,11 +29,16 @@ def annotation(self):
...
-            raise httpexceptions.HTTPNotFound()
+            raise httpexceptions.HTTPNotFound("Annotation not found")
 
         annotation_id = document["_id"]
         document_uri = document["_source"]["uri"]
 
+        if not (document_uri.startswith("http://") or
+                document_uri.startswith("https://")):
+            raise httpexceptions.HTTPUnprocessableEntity(
+                _("We can only link to http(s) URLs"))
```
chdorner: After spending some quality time sifting through all the special ways we store annotations at the moment, I found quite a few that don't have a `uri`, and even some that don't have anything that could be used as a URI. 
The Postgres migration code for example tries to read the `uri` first, and then falls back to `target.link[0]`, and if that one doesn't exist, refuses to migrate the annotation.

My suggestion for bouncer is something like:

1. Read from `uri`
2. When `uri` is missing or not a http(s) uri, find the first http(s) uri from `target.link` with `rel == 'canonical'` (careful, sometimes `target.link` is not a list but a string)
3. When that one is missing as well, show an error.

robertknight:
There are two things that could be improved here from a UX perspective:

1. Even if we cannot redirect the user to the original page, we could at least show them the content of the annotation itself, by redirecting to the standalone HTML page for an annotation or displaying the error plus a link to the annotation. The user would then at least be able to see the quote, the annotation and participate in a conversation around it.
2. The error message is short and simple but not probably not comprehensible to a lay user. As a general rule of thumb, I like errors to include: 1) A short (one-line) message which is unique to the problem and likely to be transmitted correctly via support messages, Twitter etc. 2) A short paragraph (or FAQ link) explaining the problem in laymans terms if necessary. eg. "This annotation was made on a local file. We can only show annotations in context if the annotation was made on a copy of the file with a public or private HTTP link."

bouncer/views.py
nickstenning: This is fine, but I reckon we could do with some slightly more user-friendly copy here. "Sorry, we're currently unable to link to annotations made on documents which are not publicly available."

nickstenning:
@robertknight 

1. We can't do that without bouncer acquiring a complete understanding of a) who the requesting user is, and b) the annotation permissions model. Regardless of any discussion of whether that's what we want, this is definitely out of scope at the moment.

2. Agreed. See my suggestion above.

bouncer/views.py
```diff
@@ -66,11 +75,12 @@ def index(request):
...
-    renderer='bouncer:templates/notfound.html.jinja2')
-def notfound(request):
-    request.response.status_int = 404
-    return {}
+@view.view_config(
+    context=httpexceptions.HTTPException,
+    renderer='bouncer:templates/error.html.jinja2')
+def httpexception(exc, request):
+    request.response.status_int = exc.status_int
+    return {"message": exc.message}
```
nickstenning: Does this not raise a deprecation warning? `BaseException.message` has been deprecated for a long time. Use `str(exc)` instead.

seanh: I didn't see a warning, and I think it's because Pyramid's `HTTPException` sets `self.message`. But it does also support `__str__()` as well so I'll change it to that

seanh: Done

bouncer/views.py
seanh: My suggestion here is: _"Sorry, currently we can only link to annotations of documents on the web. The annotated document must have an HTTP or HTTPS link."_ I'm trying to keep it user-friendly for non-experts (the first sentence, matches @robertknight's point 1) while also providing the necessary technical details for experts to fix the problem themselves when possible (the second sentence, matches Rob's point 2 I think), and also trying to keep it accurate (it specifically has to be an http(s) link), and concise.

conordelahunty: How about something like:

"Sorry, but it looks like this annotation was made on a document that is not publicly available.
To view it’s annotations, a document must have an HTTP or HTTPS link."

conordelahunty: Perhaps this is even clearer?

"Sorry, but it looks like this annotation was made on a document that is not publicly available.
To view it’s annotations, a document's address must start with HTTP or HTTPS."

seanh: Done

seanh:
I think in the case of non-http(s) links we could do Rob's suggestion and redirect to the annotation's standalone page, couldn't we? It would then be up to the standalone page to either show the annotation or a suitable error depending on the user's authorization. Something for another pull request perhaps

seanh:
Oh, now that I think of it we should count the number of these error responses that we send in statsd...

seanh:
Removed the double i18n of the error message in the Jinja template

seanh:
Added missing i18n of "Annotation not found"

seanh:
Might want to do the statsd reporting in a better place, so you don't have to remember to increment a statsd counter before every raise or return in a view callable. Not for this pr though

robertknight:
I think actual user testing with this scenario will be the best way to determine suitable messaging. In the absence of such feedback, the current messaging LGTM.

bouncer/views.py
```diff
@@ -25,11 +29,20 @@ def annotation(self):
...
         annotation_id = document["_id"]
         document_uri = document["_source"]["uri"]
 
+        if not (document_uri.startswith("http://") or
+                document_uri.startswith("https://")):
+            statsd.incr("views.annotation.422.not_an_http_or_https_document")
+            raise httpexceptions.HTTPUnprocessableEntity(
+                _("Sorry, but it looks like this annotation was made on a "
+                  "document that is not publicly available. "
+                  "To view it’s annotations, a document's address must start "
```
seanh: Ugh, there's a typo and a non-ASCII character in this (thanks PyLint). My fault, will fix it

nickstenning: It should be "its", anyway.

seanh: Yep

seanh: I've stuck the fix for this into https://github.com/hypothesis/bouncer/pull/11

PR #11, opened 2016-03-03
Do not crash on bad annotations

seanh:
This is branched off of <https://github.com/hypothesis/bouncer/pull/10>. Added some defensive coding against Elasticsearch document variability. I've tested this on ~100,000 annotations from the production hypothes.is API.

### DIFF
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -101,7 +101,7 @@ class AnnotationController(object):
             raise httpexceptions.HTTPUnprocessableEntity(
                 _("Sorry, but it looks like this annotation was made on a "
                   "document that is not publicly available. "
-                  "To view it’s annotations, a document's address must start "
+                  "To view its annotations, a document's address must start "
                   "with <code>http://</code> or <code>https://</code>."))
 
         via_url = "{via_base_url}/{uri}#annotations:{id}".format(
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -12,6 +12,59 @@ from statsd.defaults.env import statsd
 _ = i18n.TranslationStringFactory(__package__)
 
 
+class InvalidAnnotationError(Exception):
+
+    """Raised if an annotation from Elasticsearch can't be parsed."""
+
+    def __init__(self, message, reason):
+        """
+        Return a new InvalidAnnotationError instance.
+
+        :param message: a user-friendly error message
+        :type message: string
+
+        :param reason: a computer-friendly unique string identifying the reason
+            the exception was raised
+        :type reason: string
+
+        """
+        self.message = message
+        self.reason = reason
+
+    def __str__(self):
+        return self.message
+
+
+def parse_document(document):
+    """
+    Return the ID and URI from the given Elasticsearch annotation document.
+
+    Return the annotation ID and the annotated document's URI from the given
+    Elasticsearch annotation document.
+
+    :param document: the Elasticsearch annotation document to parse
+    :type document: dict
+
+    :rtype: 2-tuple of annotation ID (string) and document URI (string)
+
+    """
+    # We assume that Elasticsearch documents always have "_id" and "_source".
+    annotation_id = document["_id"]
+    annotation = document["_source"]
+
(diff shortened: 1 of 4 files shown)


bouncer/views.py
```diff
@@ -25,10 +30,27 @@ def annotation(self):
...
+        try:
+            annotation_id = document["_id"]
+            document_uri = document["_source"]["uri"]
+        except KeyError:
+            raise httpexceptions.HTTPUnprocessableEntity(
+                _("The annotation is invalid"))
+
+        if not isinstance(document_uri, str):
+            raise httpexceptions.HTTPUnprocessableEntity(
+                _("The annotation has an invalid document URI"))
```
chdorner: I've moved this comment over from #10, since it's more appropriate here.

After spending some quality time sifting through all the special ways we store annotations at the moment, I found quite a few that don't have a `uri`, and even some that don't have anything that could be used as a URI.
The Postgres migration code for example first tries to read the URI from `target[0][source]`, if that doesn't exist it tries to fall back to `uri`, and if taht one doesn't exist it refuses to migrate the annotation.

My suggestion for bouncer is something like:

1. Read from `target[0][source]`
2. When still missing or not a http(s) uri, try `uri`
3. Maybe: When still missing, find the first http(s) uri from `target[0].link` with `rel == 'canonical'` (careful, sometimes `target[0].link` is not a list but a string)
4. When all of the above fail, render an error

robertknight:
Is there any way we could re-use the recently written presenter logic to avoid having to deal with this variability in Bouncer itself?

nickstenning:
This looks ok, but pulling out a function `annotation_uri` which accepts an annotation object as input and returns a uri or `None` as output would make this _way_ clearer and a hell of a lot easier to test.

nickstenning:
> Is there any way we could re-use the recently written presenter logic to avoid having to deal with this variability in Bouncer itself?

As mentioned in Slack, I don't think this is worth our time. Bouncer is only reading from Elasticsearch until such time as annotation data moves to Postgres, at which point much of this won't be necessary.

seanh:
Gonna wait until #10 is merged before finishing this one off

bouncer/views.py
```diff
@@ -25,10 +30,27 @@ def annotation(self):
...
             statsd.incr("views.annotation.404.annotation_not_found")
-            raise httpexceptions.HTTPNotFound()
+            raise httpexceptions.HTTPNotFound("Annotation not found")
+
+        try:
+            annotation_id = document["_id"]
+            document_uri = document["_source"]["uri"]
+        except KeyError:
+            raise httpexceptions.HTTPUnprocessableEntity(
+                _("The annotation is invalid"))
```
seanh: This should increment a statsd counter before raising

bouncer/views.py
```diff
@@ -25,10 +30,27 @@ def annotation(self):
...
+
+        if not isinstance(document_uri, str):
+            raise httpexceptions.HTTPUnprocessableEntity(
```
seanh: This should increment a statsd counter before raising

bouncer/views.py
seanh: Done

bouncer/views.py
seanh: Done

seanh:
Done everything except @chdorner's comment about checking `["target"][0]["source"]` and `["target"][0]["link"]` as well as `["uri"] for annotated document URIs.

Split out a few utility things from `views` into a new `util` module (I wasn't going to do this, but it turned out I had a few utils not just the one function).

Addition some missing tests for the statsd reporting.

Fixed a non-ASCII typo (from a different pull request but just fixed it here).

seanh:
Still need to test this manually and maybe put in the more robust annotation parsing. Lunch now..

seanh:
As discussed in the sprint planning meeting, let's leave out the more complex annotation parsing at least for now.

I've re-tested manually and seems to work

nickstenning:
LGTM.

PR #13, opened 2016-03-04
Check chrome.runtime.lastError

### LINKED ISSUE #9: Check chrome.runtime.lastError
Rob: Looking at the JS code, I do see an issue that we're not checking `chrome.runtime.lastError` in the sendMessage callback

seanh:
If chrome.runtime.lastError is defined then log it with console.error()
and always redirect to Via instead of Chrome.

Fixes #9

### DIFF
--- a/bouncer/scripts/test/redirect-test.js
+++ b/bouncer/scripts/test/redirect-test.js
@@ -13,6 +13,11 @@ describe('redirect()', function () {
         })
       };
     };
+    sinon.stub(window.console, 'error');
+  });
+
+  afterEach(function () {
+    window.console.error.restore();
   });
 
   it('redirects to Via if not Chrome', function () {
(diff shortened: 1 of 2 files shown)


bouncer/scripts/test/redirect-test.js
```diff
@@ -71,6 +72,40 @@ describe('redirect()', function () {
       true);
   });
 
+  it('redirects to Via if lastError is defined', function () {
+    window.chrome = {
+      runtime: {
+        sendMessage: function (id, message, callbackFunction) {
+          callbackFunction('Hey!');
+        },
+        lastError: 'There was an error'
```
robertknight: This doesn't match the actual structure of errors returned by Chrome, which is `{message: string}`, I suggest that we use a fake error which has the same structure at least.

bouncer/scripts/test/redirect-test.js
```diff
@@ -13,6 +13,7 @@ describe('redirect()', function () {
         })
       };
     };
+    window.console.error = sinon.stub();
```
robertknight: When stubbing existing methods, it is usually better to use `sinon.stub(object, methodName)` and then use `object.method.restore()` afterwards to tear down the stub. Otherwise you can end up overwriting other customizations to the functions that are performed by the test harness.

seanh:
@robertknight Both done

robertknight:
@seanh - I suggest disabling JSCS for Hound, as we did for H - since the same problem is happening and then rebasing on master.

robertknight:
LGTM. @seanh - Now that I think about it, the error that we check for here is exactly the sort of thing that would be interesting to capture with Sentry. I wonder if `navigator.sendBeacon` could be used to report the error (using the Sentry API directly) without interrupting the navigation.

PR #16, opened 2016-03-04
Use gliderlabs/alpine as base Docker image

### LINKED ISSUE #14: Use the same Docker base image as h
We want to get away from the `frolvlad/alpine-python3` Docker base image just because it's a random third-party image that we don't use anywhere else. The problem is how to get an Alpine Linux with Python 3, including a version of pip that works with Python 3. Nick pointed out how smokey does it: https://github.com/hypothesis/smokey/blob/master/smokey/Dockerfile#L4

chdorner:
And install pip via the official script, instead of getting the
python-pip package which installs Python 2.

Fixes #14

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM frolvlad/alpine-python3
+FROM gliderlabs/alpine:3.3
 MAINTAINER Hypothes.is Project and contributors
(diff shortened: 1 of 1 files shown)


seanh:
Thanks!

PR #19, opened 2016-03-09
Serve static URLs with same protocol as the site

seanh:
The CSS in the bouncer production deployment isn't being loaded because
of a mixed-content warning. Pyramid generates http:// URLs for static
assets, and bouncer is deployed with https://.

To make Pyramid generate static URLs with the same protocol as the
request you need to do:

    config.add_static_view(name="//htp.is/static", path="static")

But we can't hardcode "//htp.is/static" (this differs between our stage
and production deployments, for one thing) so make it configurable by a
new environment variable.

### DIFF
--- a/bouncer/__init__.py
+++ b/bouncer/__init__.py
@@ -37,7 +37,7 @@ def app():
     config.add_static_view(name="static", path="static")
     config.include("pyramid_jinja2")
     config.registry.settings["jinja2.filters"] = {
-        "static_url": "pyramid_jinja2.filters:static_url_filter"
+        "static_path": "pyramid_jinja2.filters:static_path_filter"
     }
     config.include("bouncer.views")
     config.include("bouncer.sentry")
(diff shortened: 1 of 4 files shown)


nickstenning:
Two questions about this:

1. The nginx config for bouncer includes `include proxy_params;` which should ensure that `X-Forwarded-Proto: https` is set when requests arrive over HTTPS. Locally, this results in static URLs which start `https://`:

        $ curl -H 'X-Forwarded-Proto: https' localhost:8000
        ...
        <link rel="stylesheet"
              type="text/css"
              href="https://localhost:8000/static/styles/bouncer.css">

  Do we understand why that's not working in production?

2. Is there any reason we couldn't avoid this issue entirely by using `static_path` rather than `static_url`?

seanh:
Ok, you're right `X-Forwarded-Proto` seems to work in dev for me too, I don't know why it's not working in prod. I'd have to setup nginx talking to bouncer running in gunicorn in docker on a vm to try and reproduce it, which is time consuming.

I've changed it to use `static_path` instead as that seems like a quicker fix

nickstenning:
:+1:

PR #20, opened 2016-03-10
Use the CHROME_EXTENSION_ID environment variable

seanh:
The Chrome extension ID that bouncer tries to talk to was hardcoded (to
my dev extension). Instead actually read the CHROME_EXTENSION_ID
environment variable, send it to the JavaScript code, and have the
JavaScript code use it.

### DIFF
--- a/bouncer/scripts/test/redirect-test.js
+++ b/bouncer/scripts/test/redirect-test.js
@@ -8,6 +8,7 @@ describe('redirect()', function () {
     window.document.querySelector = function () {
       return {
         textContent: JSON.stringify({
+          chromeExtensionId: 'test-extension-id',
           extensionUrl: 'http://www.example.com/example.html#annotations:AVLlVTs1f9G3pW-EYc6q',
           viaUrl: 'https://via.hypothes.is/http://www.example.com/example.html#annotations:AVLlVTs1f9G3pW-EYc6q',
         })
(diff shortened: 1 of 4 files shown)


bouncer/scripts/test/redirect-test.js
```diff
@@ -110,6 +111,24 @@ describe('redirect()', function () {
...
+  it('calls chrome.runtime.sendMessage correctly', function () {
+    window.chrome = {
+      runtime: {
+        sendMessage: sinon.stub()
+      }
+    };
+
+    redirect(function () {});
+
+    assert.equal(window.chrome.runtime.sendMessage.calledOnce, true);
```
robertknight: Not going to stop merge, but I'll note that you can make this somewhat less verbose using `assert.calledWith(spy, matchers...)`

seanh: The problem is the third argument is an inline function from `redirect.js` though

seanh: Oh I see you could use it to test just the first two though

robertknight:
LGTM.

bouncer/scripts/test/redirect-test.js
robertknight: @seanh - Sinon has a concept called [matchers](http://sinonjs.org/docs/#sinon-match-api) which allow you to provide arguments to `assert.calledWith` that are not the exact object you want to match but instead a predicate that tests whether the argument matches some condition.

PR #21, opened 2016-03-10
Fix the front page

seanh:
```
Fix exception handling

1. Fix the front page.

   The front page raises an HTTPFound which is supposed to redirect the
   browser, but which was being caught by the custom exception view
   (which was catching all HTTPExceptions) and turned into an error page.

   We no longer have a conteext=httpexceptions.HTTPException view so the
   redirect now works as intended.

2. Don't report HTTPRedirection (3xx) and HTTPNotFound (404) to Sentry.

   These were being caught by the Raven tween and reported to Sentry.

   The Raven tween is now gone, replaced by an exception view (see 3
   below) and we have an
   @view.view_config(context=httpexceptions.HTTPError)
   @view.view_config(context=httpexceptions.HTTPServerError)
   view that catches deliberately raised Pyramid exceptions and shows
   the user an error page but does not report them to Sentry.

3. Show a nice error page for non-httpexceptions, and report them to
   Sentry.

   If a bug in the code raises an arbitrary uncaught exception
   (DivisionByZero, AssertionError, ...) this is now caught by the new
   @view.view_config(context=Exception) and reported to Sentry,
   and then a generic error page is shown to the user.

   Even though HTTPException is an Exception subclass,
   context=Exception doesn't catch it

   We don't want to reveal bugs in the production code by showing
   details of these exceptions to users, but we do want to report the
   details to Sentry for ourselves.

4. If the environment variable DEBUG is set then print tracebacks to the
   terminal instead of reporting to Sentry and showing a generic error
   page.
```

### DIFF
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -80,12 +80,40 @@ def index(request):
         location=request.registry.settings["hypothesis_url"])
 
 
-@view.view_config(
-    context=httpexceptions.HTTPException,
-    renderer='bouncer:templates/error.html.jinja2')
-def httpexception(exc, request):
-    request.response.status_int = exc.status_int
-    return {"message": str(exc)}
+@view.view_defaults(renderer='bouncer:templates/error.html.jinja2')
+class ErrorController(object):
+
+    def __init__(self, exc, request):
+        self.exc = exc
+        self.request = request
+
+    @view.view_config(context=httpexceptions.HTTPError)
+    @view.view_config(context=httpexceptions.HTTPServerError)
+    def httperror(self):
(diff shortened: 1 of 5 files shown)


bouncer/views.py
```diff
@@ -80,8 +80,12 @@ def index(request):
         location=request.registry.settings["hypothesis_url"])
 
 
+@view.view_defaults(renderer='bouncer:templates/error.html.jinja2')
```
seanh: Ugh, this is wrong `@view_defaults` doesn't work here

nickstenning:
LGTM. Waiting on Travis.

PR #22, opened 2016-04-04
Display document url when redirecting from bouncer.

sheetaluk:
Currently the text under the spinner and the page title say
"redirecting." Change that to "Loading annotation for <url/path>"
so that the user has a better idea of what is hapening and what
page he can expect to be redirected to.

### DIFF
--- a/bouncer/templates/annotation.html.jinja2
+++ b/bouncer/templates/annotation.html.jinja2
@@ -1,17 +1,17 @@
 {% extends "templates/base.html.jinja2" %}
 
-{% set title = _("Hypothesis: Redirecting you&hellip;") %}
+{% set title = _("Hypothesis: Annotation for %(site)s" )|format(site=pretty_url)  %}
 
 {% block content %}
   <div class="center spinner__stationary-ring">
-    <img alt="Redirecting"
+    <img alt="{% trans %}Loading annotation for {{ pretty_url }}{% endtrans %}"
           class="center spinner__icon"
           height="24"
           src="{{'bouncer:static/images/hypothesis-icon.svg'|static_path}}"
           width="28">
     <div class="spinner__moving-ring"></div>
   </div>
-  <p class="spinner__text">{% trans %}Redirecting{% endtrans %}</p>
+  <p class="spinner__text">{% trans %}Loading annotation for {{ pretty_url }}{% endtrans %}</p>
 {% endblock %}
 
 {% block scripts %}
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -13,6 +13,12 @@ from bouncer import util
 _ = i18n.TranslationStringFactory(__package__)
 
 
+#: The maximum length that the "netloc" (the www.example.com part in
+#: http://www.example.com/example) can be in the pretty URL that is displayed
+#: to the user before it gets truncated.
+NETLOC_MAX_LENGTH = 20
+
+
 @view.view_defaults(renderer="bouncer:templates/annotation.html.jinja2")
 class AnnotationController(object):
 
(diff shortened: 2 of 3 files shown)


bouncer/templates/annotation.html.jinja2
```diff
@@ -1,17 +1,17 @@
 {% extends "templates/base.html.jinja2" %}
 
-{% set title = _("Hypothesis: Redirecting you&hellip;") %}
+{% set title = _("Hypothesis: Annotation for " ) + pretty_url  %}
```
seanh: I think this actually needs to be something like:

```jinja2
{% set title = _("Hypothesis: Annotation for %(site)s" )|format(site=pretty_url)  %}
```

The reason is that in some languages the URL might not appear at the end of that sentence, so `%(site)s` needs to be part of the string that gets passed to `_()`, so that translators can move it around. (Unfortunately this also gives translators the chance to break things by removing or changing the `%(site)s`, but I think it's what we have to do.

bouncer/templates/annotation.html.jinja2
```diff
@@ -1,17 +1,17 @@
...
 
 {% block content %}
   <div class="center spinner__stationary-ring">
-    <img alt="Redirecting"
+    <img alt="Loading annotation for {{ pretty_url }}"
```
seanh: I see that we weren't doing it before, but this one should be sent through `_()` as well

bouncer/templates/annotation.html.jinja2
```diff
@@ -1,17 +1,17 @@
...
           class="center spinner__icon"
           height="24"
           src="{{'bouncer:static/images/hypothesis-icon.svg'|static_path}}"
           width="28">
     <div class="spinner__moving-ring"></div>
   </div>
-  <p class="spinner__text">{% trans %}Redirecting{% endtrans %}</p>
+  <p class="spinner__text">{% trans %}Loading annotation for {% endtrans %}{{ pretty_url }}</p>
```
seanh: Again `pretty_url` needs to be part of the string that gets passed to the translator so they can move it around in the sentence if necessary. On this line we're using the `{% trans %}` tag so I think you can actually just do this:

```jinja2
{% trans %}Loading annotation for {{ pretty_url }}{% endtrans %}
```

bouncer/views.py
```diff
@@ -60,6 +60,11 @@ def annotation(self):
         extension_url = "{uri}#annotations:{id}".format(
             uri=document_uri, id=annotation_id)
 
+        parsed_url = parse.urlparse(document_uri)
+        parsed_path = parsed_url.path if parsed_url.path != "/" else ""
+        if parsed_path.endswith("/"): parsed_path = parsed_path[:-1]
```
seanh: Since we're now stripping the trailing slash on line 65, line 64 can just be `parsed_path = parsed_url.path`

seanh:
Looks good! Just some small suggestions.

seanh:
By the way for i18n this is a good guide: <http://docs.ckan.org/en/latest/contributing/string-i18n.html> Some of it is CKAN or Pylons-specific but a lot of it is just covering i18n in Jinja2.

sheetaluk:
Thanks @seanh , fixed the comments please PR again :)

bouncer/views.py
```diff
@@ -60,6 +60,11 @@ def annotation(self):
         extension_url = "{uri}#annotations:{id}".format(
             uri=document_uri, id=annotation_id)
 
+        parsed_url = parse.urlparse(document_uri)
+        parsed_path = parsed_url.path
+        if parsed_path.endswith("/"): parsed_path = parsed_path[:-1]
+        pretty_url = parsed_url.netloc + parsed_path
```
seanh: Do you want to add `parsed_path` on here, actually? This produces "Loading annotation for www.seanh.cc/posts/functional-core-imperative-shell" for example. Do we just want "Loading annotation for www.seanh.cc" instead?

seanh:
@sheetaluk I added a couple of commits to this, looks good

nickstenning:
I think adding the URL is great, but I'm not sure about the copy change from "Redirecting..." to "Loading annotation for..."

We're not loading the annotation. We're redirecting you. In addition, "loading annotation" makes it sound to the user like any slowness they might experience is our fault, whereas in the case where they have the extension installed this is definitely not true.

Perhaps we could have a quick conflab about this tomorrow before deploying it?

PR #26, opened 2016-04-21
Fix typo.

sheetaluk:


chdorner:
LGTM. Tiny nitpick is that there are some stray `"` at the end of the CHANGES file, it's not going to keep me back from merging this though :)

PR #29, opened 2016-05-26
Jenkins

chdorner:
Configure Jenkins auto builds like we have with h.

This also introduces the use of `tox` for easier test-running inside the docker container on Jenkins.


nickstenning:
LGTM.

PR #30, opened 2016-06-09
Redirect PDF annotations

chdorner:
Fall back to http document link when PDF annotation
    
We now store the PDF fingerprint in the target source for annotations
made on PDF files. To be able to still redirect the user to the PDF, we
try and load an http/https URL from the document's links.

### DIFF
--- a/bouncer/test/util_test.py
+++ b/bouncer/test/util_test.py
@@ -37,3 +37,30 @@ def test_parse_document_returns_document_uri():
     })[1]
 
     assert document_uri == "http://example.com/example.html"
+
+def test_parse_document_returns_document_uri_from_links_when_pdf():
+    document_uri = util.parse_document({
+        "_id": "annotation_id",
+        "_source": {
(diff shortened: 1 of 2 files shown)


bouncer/test/util_test.py
```diff
@@ -33,7 +33,38 @@ def test_parse_document_returns_annotation_id():
...
+    assert document_uri == "http://example.com/foo.pdf"
+
+
+def test_parse_document_raises_when_uri_from_links_not_string_for_pdfs():
+    with pytest.raises(util.InvalidAnnotationError) as exc:
+        util.parse_document({
+            "_id": "annotation_id",
+            "_source": {"target": [{"source": 52}]}  # "uri" isn't a string.
+        })
+        util.parse_document({
```
nickstenning: I think you're missing another `with pytest.raises` line here.

chdorner: urgh, too fast for my own good. Fixed now.

nickstenning:
LGTM.

PR #32, opened 2016-09-15
Read web link from new `web_uri` when a URN is found

chdorner:
This relies on https://github.com/hypothesis/h/pull/3869 being merged,
deployed to production, and a full reindex being run. This will also fix
the issue that is tracked here: https://github.com/hypothesis/h/issues/3867


dwhly:
👏  Yay! Thank you!

PR #33, opened 2016-12-15
Extract Elasticsearch client into own module

nickstenning:
This commit extracts the Elasticsearch client (and its provision on the request object) into its own module in preparation for additional changes to the client to include: TLS support, and possibly AWS Auth (signed requests) so that we don't need to whitelist IPs on our Elasticsearch cluster.

### DIFF
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -32,7 +32,7 @@ class AnnotationController(object):
         settings = self.request.registry.settings
 
         try:
-            document = util.elasticsearch_client(settings).get(
+            document = self.request.es.get(
                 index=settings["elasticsearch_index"],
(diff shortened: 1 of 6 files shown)


bouncer/views.py
```diff
@@ -32,7 +32,7 @@ def annotation(self):
         settings = self.request.registry.settings
 
         try:
-            document = util.elasticsearch_client(settings).get(
+            document = self.request.es.get(
```
chdorner: `bouncer.util.elasticsearch_client` is not used anymore now and should be deleted.

nickstenning: Uh. FAIL. Sorry.

PR #35, opened 2016-12-16
Add a healthcheck endpoint

nickstenning:
This is necessary for Skyliner, but frankly it's probably useful regardless, as
we can configure it to be used by the ALB ourselves if need be.

### DIFF
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -20,6 +20,10 @@ _ = i18n.TranslationStringFactory(__package__)
 NETLOC_MAX_LENGTH = 20
 
 
+class FailedHealthcheck(Exception):
+    """An exception raised when the healthcheck fails."""
+
+
 @view.view_defaults(renderer="bouncer:templates/annotation.html.jinja2")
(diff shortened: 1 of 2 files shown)


bouncer/views.py
```diff
@@ -127,8 +131,21 @@ def error(self):
...
+    except exceptions.ElasticsearchException as exc:
+        raise FailedHealthcheck('elasticsearch exception') from exc
+    if status not in ('yellow', 'green'):
+        raise FailedHealthcheck('cluster status was {!r}'.format(status))
+    return 'imok\n'
+
+
 def includeme(config):
     config.add_route("index", "/")
+    config.add_route("healthcheck", "/ruok")
```
chdorner: Why not just something more conventional, like `/status` or `/health`?

nickstenning: Because it's fun! 😛 

(Seriously, why risk conflicting with an actually useful URL path that we might use for something user-facing in the future...?)

PR #36, opened 2017-02-09
Convert bouncer to an application

chdorner:
Some of the upcoming changes include adding multiple config files (supervisor,
collectd), and it doesn't really make sense to keep bouncer as a Python
package anymore in my opinion.

This also means that we won't have specific versions anymore but have
the same approach to version numbers as h, namely the date and sha of
the most recent commit. This will also be reflected in the version
number that we report to Sentry.


nickstenning:
Yay! Long overdue. Very nice.

PR #37, opened 2017-02-09
Docker updates

chdorner:
~~This depends on #36 and needs to be rebased after #36 has been merged.~~

---

This updates the Docker base image to alpine 3.4, bringing it to the same version as we use with h.

PR #38, opened 2017-02-09
Use supervisor to run gunicorn

chdorner:
~~This depends on #37 and needs to be rebased after #37 has been merged.~~

---

This will allow us to run multiple processes in the same container. One
of the next changes will be to run collectd inside the container, in
order to get system metrics on Skyliner.


nickstenning:
LGTM.

PR #40, opened 2017-02-10
Wrap supervisor into an `init-env` script

chdorner:
Which sets up some environment variables, most importantly it adds the
EC2 instance id (or default to `hostname`) at the end of
`STATSD_PREFIX`.


nickstenning:
:ok_hand:

PR #42, opened 2017-04-13
Bypass for bouncer to accept filter requests.

greebie:
This pull request is a prototype to enable query requests through the hyp.is server, just to show a working piece of code related to the changes mentioned [here] (https://github.com/hypothesis/client/pull/295).

This assumes a filter query syntax using query:keywords or q:keywords.

No unit tests have been created (test cases considered are simply a query bouncing to the appropriate urls and producing an error).

Collecting statistics for search queries could be considered as well.

### DIFF
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -7,6 +7,7 @@ from pyramid import httpexceptions
 from pyramid import i18n
 from pyramid import view
 from statsd.defaults.env import statsd
+import re
 
 from bouncer import util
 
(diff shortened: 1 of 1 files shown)


nickstenning:
Hi Ryan! I've taken a look at this, the Chrome extension PR, and the more general proposal which Robert sent around, and we've been discussing it amongst ourselves out here in SF.

We'd like to propose an alternate way forward which involves a little more work on our part, but which probably provides a better user experience. Namely: to give bouncer a URL-shortener-like API which can be used to generate shortlinks to a particular `(URL, client configuration)` tuple. This would mean that the resulting URLs could look very similar to the existing bouncer URLs, e.g.

```
https://hyp.is/6AR3ZZ4i/www.nytimes.com/2017/05/06/us/...
```

This would have a couple of advantages:

1. We'd be reducing the risk of introducing URL encoding/double-encoding problems.
2. The URL slug on the end is for display purposes only, so we'd be reducing the risk of the link being broken by (e.g.) line-breaking when shared by email.
3. The complexity of the query/client state configuration doesn't all have to be put in the URL.

This probably isn't a particularly clear articulation of what I'm proposing, but I'll write this up more clearly in the coming weeks and update @robertknight's thread on the developer mailing list.

greebie:
Thanks Nick.  I was aware that some broader scope changes would be necessary, so I look forward to seeing the new approach!  Thanks!

nickstenning:
I've followed up [here on the dev list](https://groups.google.com/a/list.hypothes.is/d/msg/dev/sFH-3xmeK8Y/NLIBlgb_BQAJ).

Closing the PR for now. Hope that's ok with you.

PR #43, opened 2017-05-30
use quote and text (if available) to enhance share card

judell:
# sample results

image

# see also

https://docs.google.com/document/d/1UPPPgLLPMTwxh83BITCiSK7QxfwpORONiZF9vFL4doA

# notes

these flake8 complaints were for the pre-existing HEAD (ce16736e40222dc0caab041f07ac2c6a53e4ed35)

```
./views.py:75:11: E111 indentation is not a multiple of four
./views.py:140:64: E999 SyntaxError: invalid syntax
./__init__.py:45:80: E501 line too long (97 > 79 characters)
./__init__.py:47:80: E501 line too long (105 > 79 characters)
./__init__.py:49:80: E501 line too long (83 > 79 characters)
./_version.py:62:44: F821 undefined name 'FileNotFoundError'
```

### DIFF
--- a/bouncer/templates/annotation.html.jinja2
+++ b/bouncer/templates/annotation.html.jinja2
@@ -3,6 +3,9 @@
 {% set title = _("Hypothesis: Annotation for %(site)s" )|format(site=pretty_url)  %}
 {% set meta_title = _("Annotation for %(site)s" )|format(site=pretty_url)  %}
 
+{% set quote = quote  %}
+{% set text = text  %}
+
 {% block content %}
   <div class="center spinner__stationary-ring">
     <img alt="{% trans %}Loading annotation for {{ pretty_url }}{% endtrans %}"
--- a/bouncer/templates/annotation.html.jinja2
+++ b/bouncer/templates/annotation.html.jinja2
@@ -1,8 +1,5 @@
 {% extends "templates/base.html.jinja2" %}
 
-{% set title = _("Hypothesis: Annotation for %(site)s" )|format(site=pretty_url)  %}
-{% set meta_title = _("Annotation for %(site)s" )|format(site=pretty_url)  %}
-
 {% block content %}
   <div class="center spinner__stationary-ring">
     <img alt="{% trans %}Loading annotation for {{ pretty_url }}{% endtrans %}"
--- a/bouncer/templates/annotation.html.jinja2
+++ b/bouncer/templates/annotation.html.jinja2
@@ -3,9 +3,6 @@
 {% set title = _("Hypothesis: Annotation for %(site)s" )|format(site=pretty_url)  %}
 {% set meta_title = _("Annotation for %(site)s" )|format(site=pretty_url)  %}
 
-{% set quote = quote  %}
-{% set text = text  %}
-
 {% block content %}
   <div class="center spinner__stationary-ring">
     <img alt="{% trans %}Loading annotation for {{ pretty_url }}{% endtrans %}"
--- a/bouncer/templates/base.html.jinja2
+++ b/bouncer/templates/base.html.jinja2
@@ -3,7 +3,7 @@
   <head>
     <meta charset=utf-8>
 
-    {% if visibility != 'private' %}
+    {% if shared and group == '__world__'  %}
 
     <meta name="twitter:card" content="summary" />
     <meta name="twitter:site" content="@hypothes_is" />
--- a/bouncer/templates/base.html.jinja2
+++ b/bouncer/templates/base.html.jinja2
@@ -5,14 +5,14 @@
 
     <meta name="twitter:card" content="summary" />
     <meta name="twitter:site" content="@hypothes_is" />
-    <meta name="twitter:title" content="{{ meta_title|safe }}" />
-    <meta name="twitter:description" content="Follow the link to see this annotation on the original page.  Use Hypothesis to annotate the Web and PDFs." />
+    <meta name="twitter:title" content="{{ quote }}" />
+    <meta name="twitter:description" content="{{ text }}" />
     <meta name="twitter:image" content="{{ 'bouncer:static/images/twitter.png'|static_url }}" />
 
     <meta property="og:type" content="article"/>
     <meta property="og:site_name" content="Hypothes.is"/>
-    <meta property="og:title" content="Check out this annotation."/>
-    <meta property="og:description" content="Follow the link to see this annotation on the original page. Use Hypothesis to annotate the Web and PDFs."/>
+    <meta property="og:title" content="{{ quote }}"/>
+    <meta property="og:description" content="{{ text }}"/>
     <meta property="og:image" content="{{'bouncer:static/images/facebook.png'|static_url }}"/>
 
     <title>{{ title|safe }}</title>
--- a/bouncer/templates/base.html.jinja2
+++ b/bouncer/templates/base.html.jinja2
@@ -31,8 +31,4 @@
   </body>
   {% block scripts %}{% endblock %}
 
-<div> can_reveal_metadata {{ can_reveal_metadata }}</div>
-<div>quote {{ quote }}</div>
-<div>text {{ text }}</div>
-<div>title {{ title }}</div>
 </html>
--- a/bouncer/templates/base.html.jinja2
+++ b/bouncer/templates/base.html.jinja2
@@ -3,6 +3,8 @@
   <head>
     <meta charset=utf-8>
 
+    {% if visibility != 'private' %}
+
     <meta name="twitter:card" content="summary" />
     <meta name="twitter:site" content="@hypothes_is" />
     <meta name="twitter:title" content="{{ quote }}" />
@@ -14,6 +16,8 @@
     <meta property="og:title" content="{{ quote }}"/>
     <meta property="og:description" content="{{ text }}"/>
     <meta property="og:image" content="{{'bouncer:static/images/facebook.png'|static_url }}"/>
+
+    {% endif %}
 
     <title>{{ title|safe }}</title>
     <link rel="stylesheet"
--- a/bouncer/templates/base.html.jinja2
+++ b/bouncer/templates/base.html.jinja2
@@ -3,7 +3,7 @@
   <head>
     <meta charset=utf-8>
 
-    {% if shared and group == '__world__'  %}
+    {% if can_reveal_metadata == True %}
 
     <meta name="twitter:card" content="summary" />
     <meta name="twitter:site" content="@hypothes_is" />
@@ -30,4 +30,9 @@
     </div>
   </body>
   {% block scripts %}{% endblock %}
+
+<div> can_reveal_metadata {{ can_reveal_metadata }}</div>
+<div>quote {{ quote }}</div>
+<div>text {{ text }}</div>
+<div>title {{ title }}</div>
 </html>
--- a/bouncer/templates/base.html.jinja2
+++ b/bouncer/templates/base.html.jinja2
@@ -3,7 +3,7 @@
   <head>
     <meta charset=utf-8>
 
-    {% if can_reveal_metadata == True %}
+    {% if show_metadata == True %}
 
     <meta name="twitter:card" content="summary" />
     <meta name="twitter:site" content="@hypothes_is" />
--- a/bouncer/util.py
+++ b/bouncer/util.py
@@ -29,13 +29,11 @@ class InvalidAnnotationError(Exception):
 
 def parse_document(document):
     """
-    Return the ID, URI, quote, and text from the given Elasticsearch
-    annotation document.
+    Return the annotation ID, annotated document's URI, and shared
+    status from the given Elasticsearch annotation document.
 
-    Return the annotation ID and the annotated document's URI from the given
-    Elasticsearch annotation document.
-
-    Also return quote (if available) and text to enhance the share card.
+    Also return annotation quote (if available, else empty) and text
+    to enhance the share card.
 
     :param document: the Elasticsearch annotation document to parse
     :type document: dict
@@ -44,29 +42,33 @@ def parse_document(document):
 
     """
     # We assume that Elasticsearch documents always have "_id" and "_source".
-
     annotation_id = document["_id"]
     annotation = document["_source"]
-    quote = ""
-    text = ""
+
+    # And that annotations always have "group" and "shared"
+    group = annotation["group"]
+    shared = annotation["shared"]
+
+    text = annotation.get('text', 'Follow this link to see the annotation on the original page.')
 
     document_uri = None
-
-    if "text" not in annotation:
-        text = "Follow this link to see the annotation on the original page."
-    else:
-        text = annotation["text"]
+    quote = None
 
     try:
         targets = annotation["target"]
         if targets:
             document_uri = targets[0]["source"]
-            selectors = targets[0]["selector"]
-            for selector in selectors:
-                if selector["type"] == "TextQuoteSelector":
-                    quote = selector["exact"]
+            if 'selector' in targets[0]:
+                selectors = targets[0]["selector"]
+                for selector in selectors:
+                    if "type" in selector and selector["type"] == "TextQuoteSelector":
+                        if "exact" in selector:
+                            quote = selector["exact"]
     except KeyError:
         pass
+
+    if quote is None:
+        quote = "Annotation for {}".format(document_uri)
 
     if isinstance(document_uri, str) and document_uri.startswith("urn:x-pdf:"):
         try:
@@ -76,16 +78,24 @@ def parse_document(document):
         except KeyError:
             pass
 
-    if quote == "":
-        quote = "Annotation for {}".format(document_uri)
-
     if document_uri is None:
         raise InvalidAnnotationError(
             _("The annotation has no URI"), "annotation_has_no_uri")
+
+    if quote is None:
+        raise InvalidAnnotationError(
+            _("The annotation has a TextQuoteSelector but no exact quote"), "annotation_has_no_quote")  
 
     if not isinstance(document_uri, str):
         raise InvalidAnnotationError(
             _("The annotation has an invalid document URI"),
             "uri_not_a_string")
 
-    return (annotation_id, document_uri, quote, text)
+    return {
+            "annotation_id": annotation_id,
+            "document_uri": document_uri,
+            "group": group,
+            "shared": shared,
+            "quote": quote,
+            "text": text
+            }
--- a/bouncer/util.py
+++ b/bouncer/util.py
@@ -29,10 +29,13 @@ class InvalidAnnotationError(Exception):
 
 def parse_document(document):
     """
-    Return the ID and URI from the given Elasticsearch annotation document.
+    Return the ID, URI, quote, and text from the given Elasticsearch
+    annotation document.
 
     Return the annotation ID and the annotated document's URI from the given
     Elasticsearch annotation document.
+
+    Also return quote (if available) and text to enhance the share card.
 
     :param document: the Elasticsearch annotation document to parse
     :type document: dict
@@ -41,15 +44,26 @@ def parse_document(document):
 
     """
     # We assume that Elasticsearch documents always have "_id" and "_source".
+
     annotation_id = document["_id"]
     annotation = document["_source"]
 
     document_uri = None
 
+    quote = None
+    if annotation["text"] == "":
+        text = "Follow this link to see the annotation on the original page."
+    else:
+        text = annotation["text"]
+
     try:
         targets = annotation["target"]
         if targets:
             document_uri = targets[0]["source"]
+            selectors = targets[0]["selector"]
+            for selector in selectors:
+                if selector["type"] == "TextQuoteSelector":
+                    quote = selector["exact"]
     except KeyError:
         pass
 
@@ -61,6 +75,9 @@ def parse_document(document):
         except KeyError:
             pass
 
+    if quote is None:
+        quote = "Annotation for " + document_uri
+
     if document_uri is None:
         raise InvalidAnnotationError(
             _("The annotation has no URI"), "annotation_has_no_uri")
@@ -70,4 +87,4 @@ def parse_document(document):
             _("The annotation has an invalid document URI"),
             "uri_not_a_string")
 
-    return (annotation_id, document_uri)
+    return (annotation_id, document_uri, quote, text)
--- a/bouncer/util.py
+++ b/bouncer/util.py
@@ -1,7 +1,15 @@
+from urllib import parse
 from pyramid import i18n
+import jinja2
 
 
 _ = i18n.TranslationStringFactory(__package__)
+
+
+#: The maximum length that the "netloc" (the www.example.com part in
+#: http://www.example.com/example) can be in the pretty URL that is displayed
+#: to the user before it gets truncated.
+NETLOC_MAX_LENGTH = 30
 
 
 class InvalidAnnotationError(Exception):
@@ -47,14 +55,15 @@ def parse_document(document):
 
     # And that annotations always have "group" and "shared"
     group = annotation["group"]
-    shared = annotation["shared"]
+    is_shared = annotation["shared"] is True
 
-    boilerplate_text = (
-        'Follow this link to see the annotation on the original page.')
-    text = (annotation.get('text', boilerplate_text))
+    can_reveal_metadata = is_shared and group == "__world__"
 
     document_uri = None
     quote = None
+    text = annotation.get('text', make_boilerplate_text())
+    if text == "":
+        text = make_boilerplate_text()
 
     try:
         targets = annotation["target"]
@@ -62,15 +71,18 @@ def parse_document(document):
             document_uri = targets[0]["source"]
             if 'selector' in targets[0]:
                 selectors = targets[0]["selector"]
-                for sel in selectors:
-                    if "type" in sel and sel["type"] == ("TextQuoteSelector"):
-                        if "exact" in sel:
-                            quote = sel["exact"]
+                for selector in selectors:
+                    if selector.get('type') != "TextQuoteSelector":
+                        continue
+                    quote = selector.get("exact")
     except KeyError:
         pass
 
     if quote is None:
-        quote = "Annotation for {}".format(document_uri)
+        quote = make_boilerplate_quote(document_uri)
+
+    if text is None:
+        text = make_boilerplate_text()
 
     if isinstance(document_uri, str) and document_uri.startswith("urn:x-pdf:"):
         try:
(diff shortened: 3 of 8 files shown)


chdorner:
We have to be careful here. Because bouncer does not know anything about access controls. Back when we were planning bouncer we decided that the URL will be public information, regardless of the group/shared state of an annotation. But surfacing any more information than that needs to be done carefully.

chdorner:
I probably should expand a little bit here:

I don't think we can show any other information than the URL for annotations that are either in a group that is not world readable, or is a private annotation. This is because social networks like Facebook and Twitter scrape this information and then show it to all their users. So we have no idea who will see this, meaning that we cannot make private data public.

The obvious solution is to check that the group is world readable and the annotation is publicly shared. However there are some caveats:

* Bouncer has only access to the search index and thus cannot know if the group is world readable or not (that's only in Postgres)
* Even if we could do that, there is the issue of a social network scraping and caching this information from a publicly available annotation which the user might make private after a while. This data is however still cached on Facebook/Twitter etc. and shown to users even though it shouldn't. We could argue that this is okay because it started out public, but it's something we have to consider carefully.

judell:
Excellent points. They also apply to the more advanced version of this idea ("textshots") that @sean-roberts has been working on.

One place to address them might be here:

image

The client knows I'm about to share a group-private annotation. Could it decorate the bouncer URL in some  way to convey that knowledge to bouncer?

chdorner:
> The client knows I'm about to share a group-private annotation. Could it decorate the bouncer URL in some way to convey that knowledge to bouncer?

I'm not sure I understand what you mean. The user copying the bouncer link does have access to the annotation, but users seeing the shared link (with the proposed text) might not be allowed to see that information.
Or did I misunderstand the alternative you've described?

judell:
"did I misunderstand the alternative you've described?"

@chdorner, I circled back to this today. Here's what I had in  mind. 

In client/src/sidebar/templates/annotation-share-dialog.html:

     <input class="annotation-share-dialog-link__text"
-      value="{{vm.uri}}" readonly>
+      value="{{vm.uri}}?private={{vm.isPrivate}}" readonly>

In bouncer/views.py:

+        if ( 'private' in self.request.GET.keys() and
+              self.request.GET['private'] == 'true' ):
+            quote = 'private'
+            text = 'private'

Do you think something like this will properly protect the content?

sean-roberts:
I feel that having "Private" be on the values is overkill. I'd say we should fallback to having no meta tag upgrades in this case

judell:
"having no meta tag upgrades in this case"

What would that look like?

sean-roberts:
As in having meta tags is the upgraded experience from what we have today. If the annotation is private, don't have any meta tags or descriptors.

chdorner:
> As in having meta tags is the upgraded experience from what we have today. If the annotation is private, don't have any meta tags or descriptors.

This is fine by me. We should just have a discussion about the caching behaviour of the systems that scrape these tags. Because a previously public annotation that has been cached by Facebook/Twitter/etc. that now changed to private will still be cached on these systems and showing the cards to their users.

judell:
The fb and twitter metadata is now omitted in the case of a private annotation.

> a previously public annotation that has been cached by Facebook/Twitter/etc. that now 
> changed to private will still be cached on these systems and showing the cards to their users

Would it be overkill to alert the user to that possibility when switching to private visibility?

judell:
@chdorner, @sean-roberts any further thoughts on this?

bouncer/util.py
```diff
@@ -29,27 +29,42 @@ def __str__(self):
...
     annotation = document["_source"]
+    quote = ""
+    text = ""
 
     document_uri = None
 
+    if "text" not in annotation:
+        text = "Follow this link to see the annotation on the original page."
+    else:
+        text = annotation["text"]
```
chdorner: Personal opinion, no need to change if you disagree: You can make setting the `text` variable a bit more straight forward doing:
```
text = annotation.get('text', 'Follow this link to see the annotation on the original page.')
```
And at the same time you can remove the empty assignment on line 51 (`text = ""`).

bouncer/util.py
```diff
@@ -61,6 +76,9 @@ def parse_document(document):
         except KeyError:
             pass
 
+    if quote == "":
+        quote = "Annotation for {}".format(document_uri)
```
chdorner: Do we need a to show a text if there is no exact quote selector? This wouldn't look so great for a complicated url: "Annotation for https://www.cambridge.org/core/books/history-manifesto/big-questions-big-data/F60D7E21EFBD018F5410FB315FBA4590/core-reader"

Depending on how Facebook and Twitter render the cards, it might look nicer to just not add the `description` meta tag in this case?

bouncer/util.py
```diff
@@ -70,4 +88,4 @@ def parse_document(document):
             _("The annotation has an invalid document URI"),
             "uri_not_a_string")
 
-    return (annotation_id, document_uri)
+    return (annotation_id, document_uri, quote, text)
```
chdorner: I think the return value is getting a bit out of hand at this point. I'd advocate using either a [`namedtuple`](https://docs.python.org/3/library/collections.html#collections.namedtuple) or a map with named keys.

bouncer/util.py
chdorner: We need to test the changes to this function as well, especially the error paths making sure that it behaves as intended when data is missing in the search index (plus the happy path as well).

bouncer/views.py
```diff
@@ -75,6 +77,12 @@ def annotation(self):
           pretty_url = pretty_url + jinja2.Markup("&hellip;")
 
         statsd.incr("views.annotation.200.annotation_found")
+        visibility = 'public'
+        if ( 'private' in self.request.GET.keys() and 
+              self.request.GET['private'] == 'true' ):
+            visibility = 'private'
+            quote = None
+            text = None
```
chdorner: Even though we need to change how bouncer discovers that an annotation is not public, we still need to have tests where given a private annotation this view will return `None` for both the quote and the text.

judell:
Thanks for the detailed review, @chdorner!

I lazily opted for the "anything that isn't in the `__world__` group is not going to have the extra information" 
 method, did the cleanup and wrote the tests you suggested, then realized, d'oh, won't work, "Only Me" annotations can of course be in `__world__`.

So this really does seem to require a way to ask the service if an annotation is shared. I agree we ought not dig it out of ElasticSearch, so "an internal services API that does not respect any of the access rules, but is only available to select (internal) services" sounds good -- but daunting.

I wonder if [this](https://github.com/hypothesis/h/compare/annotation-shared) is safe, stylistically OK, and could maybe suffice?  

The idea is that, although in general nobody should be seeing an annotation ID who isn't authorized to view it, IDs can and will leak for various reasons. But a service that simply answers true or false to the question "Is the annotation with this ID shared?" could perhaps be openly available. In which case bouncer could use it to do this:

```
{% if shared and group == '__world__'  %}
```

I've got a version of this PR working that way, but it depends on something like this:


```
def is_shared(annotation_id, hypothesis_url):
    """  Query the (hypothetical) ID.shared API  """                                                 
    url = "%s/api/annotations/%s.shared" % ( hypothesis_url, annotation_id )
    r = requests.get(url)
    return r.text == 'true'
```

Which I'm not sure will be kosher, you're in a far better position to judge.

chdorner:
Right, so the way it is implemented over in `h` is problematic.

1. Passing the current request URL into the `AnnotationResource` and then checking it inside seems quite hacky, which is usually a good indicator that something isn't right.
2. Checking that the URL ends with `".shared"` opens up a whole bunch of security issues, for example: We allow dots (`.`) in usernames, so any request to a user profile for the username "foobar.shared" would have read access to any annotation in the system, no matter who is logged in.
3. We currently do not expose if an annotation ID exists or not if the user has no access to it. We render a `404` when either the annotation does not exist, or when the currently authenticated (or anonymous) user has no access to that specific annotation, we don't render a `403`. Your proposed code in h would change that, which is something that needs to be considered carefully.
4. This endpoint would not benefit any other implementation other than this specific bouncer use-case. This, plus the fact that it **should** ignore the ACL are two good indicators that this endpoint should only be available on the internal network and not part of the public API.

All of this being said, we actually do [store the `Annotation.shared` property in the search index](https://github.com/hypothesis/h/blob/master/h/presenters/annotation_searchindex.py#L29). So you can do the solution that only renders the meta tags for annotations that are shared and in the `__world__` group.

judell:
> we actually do store the shared property

Aha. OK, thanks, I'll come back with a solution that uses it.

Meanwhile, for my own edification...

> a good indicator that something isn't right

Yep, did not smell good to me either, wanted your views as to why.

> any request to a user profile for the username "foobar.shared" would 
> have read access to any annotation in the system

Wouldn't the route need to match /api/annotations/{id:[A-Za-z0-9_-]{20,22}}.shared ?

> We currently do not expose if an annotation ID exists or not if the user
> has no access to it

Got it.

Separately, regarding https://github.com/hypothesis/bouncer/pull/43#discussion_r126116550 ("Do we need a to show a text if there is no exact quote selector?"), here's my thinking. With a structure like this...

```
   <meta name="twitter:card" content="summary" />
    <meta name="twitter:site" content="@hypothes_is" />
    <meta name="twitter:title" content="Annotation for jonudell.net" />
    <meta name="twitter:description" content="Follow the link to see this annotation on the original page.  Use Hypothesis to annotate the Web and PDFs." />
    <meta name="twitter:image" content="http://hyp.is/static/images/twitter.png" />

    <meta property="og:type" content="article"/>
    <meta property="og:site_name" content="Hypothes.is"/>
    <meta property="og:title" content="Check out this annotation."/>
    <meta property="og:description" content="Follow the link to see this annotation on the original page. Use Hypothesis to annotate the Web and PDFs."/>
    <meta property="og:image" content="http://hyp.is/static/images/facebook.png"/>
```

... we are filling two slots, Title (most important), and Description (less important). From an editorial perspective, our data does not map perfectly to that structure. The quote isn't a perfect analog for Title, and the comment isn't a perfect analog for Description. But if we accept that it's a good-enough mapping for this purpose, then since the Title slot is most important, it ought not be empty.

Oh, also, regarding https://github.com/hypothesis/bouncer/pull/43#discussion_r126111627 ("Do we need to consider encoding?"), I checked the headers coming back from the [Python elasticsearch module](https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/index.html), it receives `'Content-Type': 'application/json; charset=UTF-8'`, which is what bouncer serves back, so I think we're OK on that front, although I'm always prepared to be made a fool of by Python encoding issues!

judell:
I think there's yet one more thing to consider. With this scheme, an annotation that's shared in a third-party namespace will produce no metadata even if the group's `readable_by` is `world`. We could perhaps revert to the status quo in that case, sending boilerplate instead of nothing?

bouncer/util.py
```diff
@@ -44,15 +45,31 @@ def parse_document(document):
     annotation_id = document["_id"]
     annotation = document["_source"]
 
+    # And that annotations always have "group" and "shared"
+    group = annotation["group"]
+    shared = annotation["shared"]
```
chdorner: Since we're dealing with potentially exposing private information maybe we can be a bit more defensive here and change the code to `shared = annotation['shared'] == True`.

If, for whatever reason, we would change the value of `shared` in the search index to a string, then we would expose private data because a non-empty string is always truthy.

bouncer/templates/base.html.jinja2
```diff
@@ -3,18 +3,22 @@
   <head>
     <meta charset=utf-8>
 
+    {% if shared and group == '__world__'  %}
```
chdorner: Usually, this would be fine, but since we're dealing with a piece of code that could potentially leak private information in case there is a bug, maybe we should think about this a bit more.

The problem now is that you're reading the `shared` and `group` property in the view code (or code that is called by the view code) and you're returning the value to the template. The template then interprets the value and either shows the quotes or the text.

I think it would be better to only have either:

a) Only have one boolean that signals if we want to show the quote and text or not, that boolean should be assigned in the view, in which case we can test that we're checking for the correct values.

b) You were already talking about maybe going back to showing boilerplate data in cases where the annotation is private, in that case, we can make just render the `quote` and `text` variables in the template, and the view code tests can give us the confidence that we're rendering the correct piece of data depending on the input.

bouncer/util.py
```diff
@@ -65,9 +82,20 @@ def parse_document(document):
         raise InvalidAnnotationError(
             _("The annotation has no URI"), "annotation_has_no_uri")
 
+    if quote is None:
```
chdorner: This whole `if` block is never being called because a few lines before this you check if `quote` is `None` and if so, you give it a boilerplate value. It should be removed.

tests/bouncer/util_test.py
```diff
@@ -7,7 +7,13 @@ def test_parse_document_raises_if_no_uri():
     with pytest.raises(util.InvalidAnnotationError) as exc:
```
chdorner: There are a bunch more code style mistakes in this file (still no idea why Hound doesn't pick it up, maybe it isn't configured for this repo). Just briefly looking the output for `pep8 tests/bouncer/util_test.py` it looks like most of the errors are because of changes in this branch. Those should be cleaned up.

bouncer/util.py
```diff
@@ -44,15 +45,31 @@ def parse_document(document):
...
+    text = annotation.get('text', 'Follow this link to see the annotation on the original page.')
+
     document_uri = None
+    quote = None
 
     try:
         targets = annotation["target"]
         if targets:
             document_uri = targets[0]["source"]
+            if 'selector' in targets[0]:
```
chdorner: This is a lot of nesting here which is rather hard to read. This might be improved by doing something like:

```
selectors = targets[0].get('selector', [])
for selector in selectors:
    if selector.get('type') != 'TextQuoteSelector':
        continue
    quote = selector.get('exact')
```

bouncer/views.py
```diff
@@ -83,7 +90,11 @@ def annotation(self):
                 "viaUrl": via_url,
                 "extensionUrl": extension_url,
             }),
-            "pretty_url": pretty_url
+            "pretty_url": pretty_url,
+            "quote": quote,
+            "text": text,
+            "shared": shared,
+            "group": group
```
chdorner: There are no tests ensuring that we return the correct values for `quote`, `text`, `shared`, and `group`, but we do have tests for all the other values.

bouncer/util.py
judell: I like that better too. Although pep8 complains, and seems to prefer `is` to `==`,  I can't totally follow the arguments pro and con, so I'll use `is` for now and defer to your judgement.

bouncer/util.py
judell: Got it, thanks. So does `dict.get(key)` pretty much obsolete `dict.get[key]` at this point?

judell: > if selector.get('type') != 'TextQuoteSelector':

Somehow I never acquired that Python idiom, thank you!

judell:
So I went back to omitting the metadata for all cases but shared + world, figuring we can revisit when we have more experience with publisher groups. And there's enough complexity for now just making sure the right stuff shows up in the shared + world case, given that either or both of `quote` and `text` may be missing and require boilerplate.

> Wouldn't the route need to match /api/annotations/{id:[A-Za-z0-9_-]{20,22}}.shared ?
>> It depends.

I think I'm beginning to get the picture. In Pyramid we have a combo of the ancient Zope object traversal, plus URL dispatch, two deeply magical technologies. What could possibly go wrong? :-)

> code style mistakes 

What remain are mostly pre-existing line-too-long complaints. However at https://github.com/judell/bouncer/blob/enhanced-share-metadata/tests/bouncer/util_test.py#L153 pep8 would prefer 'is' to '==', I read a bit about pros and cons, but have no opinion.

bouncer/templates/annotation.html.jinja2
```diff
@@ -1,8 +1,5 @@
 {% extends "templates/base.html.jinja2" %}
 
-{% set title = _("Hypothesis: Annotation for %(site)s" )|format(site=pretty_url)  %}
```
chdorner: I don't think you want to remove this line, this is the HTML document title and isn't used in the meta tags.

bouncer/templates/base.html.jinja2
```diff
@@ -3,18 +3,22 @@
   <head>
     <meta charset=utf-8>
 
+    {% if can_reveal_metadata == True %}
```
chdorner: I probably messed this up in my other comments, but this should be `is True` (also with `None` and `False`), [here](https://stackoverflow.com/a/27276841)'s a StackOverflow answer providing some reasons on `is` vs `==`.

bouncer/util.py
chdorner: This still needs fixing :slightly_smiling_face:

bouncer/util.py
```diff
@@ -70,4 +101,35 @@ def parse_document(document):
...
+
+
+def make_boilerplate_quote(document_uri):
+    if document_uri is None:
+        return "Hypothesis annotation"
+    else:
+        pretty_url = make_pretty_url(document_uri)
+        return "Hypothesis annotation for {site}".format(site=pretty_url)
+
+def make_boilerplate_text():
```
chdorner: There's no real reason for this to be a function, it would be much nicer and easier to follow if this is a constant at the top of the file. Also the string should be translatable, you need to wrap it into a `_()` function call for that (e.g. `_('Follow this link ...')`)

bouncer/util.py
```diff
@@ -44,15 +53,37 @@ def parse_document(document):
...
+    group = annotation["group"]
+    is_shared = annotation["shared"] is True
+
+    can_reveal_metadata = is_shared and group == "__world__"
+
     document_uri = None
+    quote = None
+    text = annotation.get('text', make_boilerplate_text())
+    if text == "":
+        text = make_boilerplate_text()
```
chdorner: This seems to be a bit of a confusing way of setting the default multiple times. You can just do:

```
text = annotation.get('text')
if not text:
    text = ANNOTATION_BOILERPLATE_TEXT
```

This works because:
* `{'foo': 'bar'}.get('key')` returns `None` which is falsy.
* An empty string (`''`) is falsy as well.

(See comment further below for the use of a constant instead of a function)

bouncer/util.py
```diff
@@ -70,4 +101,35 @@ def parse_document(document):
...
+        parsed_url = parse.urlparse(url)
+        pretty_url = parsed_url.netloc[:NETLOC_MAX_LENGTH]
+        if len(parsed_url.netloc) > NETLOC_MAX_LENGTH:
+            pretty_url = pretty_url + jinja2.Markup("&hellip;")
+    except:
+        return "CannotParseURL"
+    return pretty_url
+
+
+def make_boilerplate_quote(document_uri):
```
chdorner: There are multiple comments about this function, I'll add them all here instead of on each line.

1. `make_` is not a very common prefix in the Python world, `get_boilerplate_quote` should be good enough
2. The passed-in `document_uri` string could be empty as well, so the `if` statement should just be `if not document_uri` (or if you don't like `not`, then just swap the if/else bodies around.
3. Both strings ("Hypothesis annotation", and "Hypothesis annotation for ...") need to be translatable so they need to be wrapped with `_()` (e.g. `_('Hypothesis annotation')`).
4. This function needs to be unit tested to ensure it behaves correctly.

bouncer/util.py
```diff
@@ -44,15 +53,37 @@ def parse_document(document):
...
+                    if selector.get('type') != "TextQuoteSelector":
+                        continue
+                    quote = selector.get("exact")
     except KeyError:
         pass
 
+    if quote is None:
+        quote = make_boilerplate_quote(document_uri)
+
+    if text is None:
```
chdorner: The `text` variable cannot be `None` at this point as you're already setting the default value a few lines above. This whole `if` block should be deleted.

bouncer/util.py
```diff
@@ -70,4 +101,35 @@ def parse_document(document):
             _("The annotation has an invalid document URI"),
             "uri_not_a_string")
 
-    return (annotation_id, document_uri)
+    return {
+            "annotation_id": annotation_id,
+            "document_uri": document_uri,
+            "can_reveal_metadata": can_reveal_metadata,
```
chdorner: No need to change, but my personal preference is to keep variable names short and descriptive. Something like `show_metadata` should be enough and you can save yourself from typing one more world followed by an underscore.

bouncer/util.py
```diff
@@ -70,4 +101,35 @@ def parse_document(document):
...
+            "quote": quote,
+            "text": text
+            }
+
+
+def make_pretty_url(url):
```
chdorner: Similar to another comment, `make_` is not a very common prefix. The old function name was good enough IMO (except for the underscore describing that it's a private function). Why not just `pretty_url`? Other good alternatives might be `get_pretty_url` or `generate_pretty_url`.

bouncer/util.py
```diff
@@ -70,4 +101,35 @@ def parse_document(document):
...
+    """
+    Return the domain name from `url` for display.
+    """
+    try:
+        parsed_url = parse.urlparse(url)
+        pretty_url = parsed_url.netloc[:NETLOC_MAX_LENGTH]
+        if len(parsed_url.netloc) > NETLOC_MAX_LENGTH:
+            pretty_url = pretty_url + jinja2.Markup("&hellip;")
+    except:
+        return "CannotParseURL"
```
chdorner: I think I wrote a comment saying that we should have error handling for this because parsing a URL can theoretically fail. I can't find that comment anymore, but there are two comments I have regarding this error handling:

1. We should avoid catching all errors (`except:`) as much as possible, this could easily mask a bug in the code that we will never notice. The [documentation for urllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) mentions that it raises `ValueError` in certain circumstances. So we should only check for `ValueError` and let any other exception blow up.

2. Returning the cryptic string "CannotParseURL" would, for example, show the string "Hypothesis annotation for CannotParseURL" to the user, which isn't really helpful. I think it would make more sense for this function to return `None` in case of a `ValueError` and then we show the normal boilerplate ("Hypothesis annotation") string in the view instead of the more detailed one.

bouncer/util.py
```diff
@@ -70,4 +101,35 @@ def parse_document(document):
...
+
+def make_pretty_url(url):
+    """
+    Return the domain name from `url` for display.
+    """
+    try:
+        parsed_url = parse.urlparse(url)
+        pretty_url = parsed_url.netloc[:NETLOC_MAX_LENGTH]
+        if len(parsed_url.netloc) > NETLOC_MAX_LENGTH:
+            pretty_url = pretty_url + jinja2.Markup("&hellip;")
```
chdorner: The `return pretty_url` a few lines below should probably be moved here at the end of the `try` block, for better readability.

bouncer/views.py
```diff
@@ -69,7 +69,12 @@ def annotation(self):
         extension_url = "{uri}#annotations:{id}".format(
             uri=document_uri, id=annotation_id)
 
-        pretty_url = _pretty_url(document_uri)
+        pretty_url = util.make_pretty_url(document_uri)
+
+        if can_reveal_metadata:
+            title = "Hypothesis annotation for {site}".format(site=pretty_url)
+        else:
+            title = "Hypothesis annotation"
```
chdorner: These two strings are already defined in functions (or constants) in the `util` package, why not just use those instead of copying them here?
They're also not translatable at the moment.

In addition to that, there doesn't seem to be any tests for both of these branches, right?

tests/bouncer/search_test.py
```diff
@@ -50,7 +50,8 @@ def test_configures_es_for_aws_auth(self, es_mock, awsauth_mock):
                                         use_ssl=True,
```
chdorner: I guess you fixed some PEP8 warnings in this file? Which is great, thanks! It's just that sometimes it makes a PR hard to review if there are unrelated changes to the problem you're trying to solve. Just a tip for the future :slightly_smiling_face:

bouncer/templates/annotation.html.jinja2
judell: It's provided in the data given to the template.

bouncer/util.py
judell: This was fixed, then regressed because of a branch mixup that happened because I'm doing this in a fork.

judell:
Thanks for your  patience, @chdorner, I really appreciate your guidance in all this. I hope the requested changes and suggestions are properly addressed, if not I will chip away until they are.

A new wrinkle here is escaping quotes in the metadata attributes, which is happening here --  https://github.com/hypothesis/bouncer/pull/43/files#diff-0dc676207c2a4bdae682669c7a3380f3R148 -- but perhaps there's a better way? I considered also removing newlines, because the services that display this metadata (well, at least Facebook and Twitter, which I've checked) will do so anyway. However https://validator.w3.org/nu/ seems happy with multiline quoted attributes so maybe better not to preempt what the downstream service?

BTW the tools for checking how FB and Twitter display share metadata are:

https://developers.facebook.com/tools/debug/sharing/

https://cards-dev.twitter.com/validator

Maybe those links should be included in comments in the code?

bouncer/templates/annotation.html.jinja2
chdorner: Good point :+1:

bouncer/templates/base.html.jinja2
```diff
@@ -3,18 +3,22 @@
   <head>
     <meta charset=utf-8>
 
+    {% if show_metadata == True %}
```
chdorner: Explicitly checking for boolean values is done with `is` instead of `==`. This should be `{% if show_metadata is True %}`.

bouncer/util.py
```diff
@@ -70,4 +113,37 @@ def parse_document(document):
...
+
+def get_boilerplate_quote(document_uri):
+    pretty_url = get_pretty_url(document_uri)
+    if pretty_url:
+        return _("Hypothesis annotation for {site}".format(site=pretty_url))
+    else:
+        return ANNOTATION_BOILERPLATE_QUOTE
+
+
+def _escape_quotes(string):
```
chdorner: Very nice :+1:

bouncer/templates/base.html.jinja2
judell: That fails like so:

jinja2.exceptions.TemplateAssertionError: no test named 'True'

According to https://github.com/pallets/jinja/issues/520#issuecomment-164306945:

> Jinja's is is different from Python's is. It invokes functions registered as "tests". 

namely: https://github.com/pallets/jinja/blob/master/jinja2/tests.py#L142

judell:
TIL that jinja's `is`  != Python's `is` ! The former, used in the template, produces the error:

> jinja2.exceptions.TemplateAssertionError: no test named 'True'

cf https://github.com/pallets/jinja/issues/520

So I guess we stay with ` {% if show_metadata == True %}`

>  Given that all of the changes are part of one logical change, maybe one commit is enough? If so, we can just "Squash and merge" the PR directly in Github and don't have to worry about messing around with git. Does this sound good to you?

Absolutely.

bouncer/templates/base.html.jinja2
chdorner: Ah yes! I always forget that jinja ships it's own small language.

PR #48, opened 2017-07-04
Add `/go` endpoint for visiting a URL with annotations.

robertknight:
Add a `/go?url={url}` endpoint which takes the user to a given URL with
the annotation layer enabled, optionally with the client configured to
initially show only annotations matching a given filter by specifying a
"q={query}" parameter.

The {query} is passed directly to the search input in the client.

For the API design rationale and discussion, see [1] and [2].

[1] https://gist.github.com/robertknight/85c9dfc89910288e84588903ab1010a3

[2] https://groups.google.com/a/list.hypothes.is/forum/#!topic/dev/sFH-3xmeK8Y

----

**Testing**

1. Visit https://BOUNCER.URL/go?url=https://example.com
2. This should open "example.com" in Via with the annotation layer enabled. Note that the sidebar does not automatically open if no query is supplied.
3. Visit https://BOUNCER.URL/go?url=https://example.com&q=robertknight
4. This should open "example.com" in Via, open the sidebar and show annotations written by or containing the string "robertknight".

Common queries would be "user:a_username" or "tag:a_tag", but be aware that there is a bug in the client where it does not correctly decode reserved URL chars (eg. ":", "/") in the query, so a query such as "user:robertknight" ends up populating the client's search input with "user%3Arobertknight". This is fixed by https://github.com/hypothesis/client/pull/483

### DIFF
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -55,7 +55,7 @@ class AnnotationController(object):
         # append our own.
         document_uri = parse.urldefrag(document_uri)[0]
 
-        if not _is_http_url(document_uri):
+        if not _is_valid_http_url(document_uri):
             statsd.incr("views.annotation.422.not_an_http_or_https_document")
             raise httpexceptions.HTTPUnprocessableEntity(
                 _("Sorry, but it looks like this annotation was made on a "
@@ -109,10 +109,10 @@ def goto_url(request):
     if url is None:
         raise httpexceptions.HTTPBadRequest('"url" parameter is missing')
 
-    if not _is_http_url(url):
+    if not _is_valid_http_url(url):
         raise httpexceptions.HTTPBadRequest(
             _('Sorry, but this service can only show annotations on '
-              'HTTP URLs.'))
+              'valid HTTP or HTTPs URLs.'))
 
     # Remove any existing #fragment identifier from the URI before we
     # append our own.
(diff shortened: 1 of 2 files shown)


bouncer/views.py
```diff
@@ -96,6 +92,54 @@ def index(request):
...
+    """
+    settings = request.registry.settings
+    url = request.params.get('url')
+
+    if url is None:
+        raise httpexceptions.HTTPBadRequest('"url" parameter is missing')
+
+    # Remove any existing #fragment identifier from the URI before we
+    # append our own.
+    url = parse.urldefrag(url)[0]
```
chdorner: Personal opinion (no need to change if you disagree): I'd put this after the `_is_http_url` call, so first you have two code blocks validating the url (not empty, http(s)), and then you start messing around with it before returning the result to the browser.

dwhly:
@robertknight Is support for tag: and group: included?  Just glanced over the code but didn't see it...

robertknight:
> @robertknight Is support for tag: and group: included? Just glanced over the code but didn't see it...

The "q" param populates the client's search field so this supports all the fields that the client's search input supports, which means:

`tag:{tag}` - Yes.
`group:{pubid}` - No, not yet. I do understand that this would be very useful. It does surface some issues that we'll need to address or have a reasonable idea of what we might do later, such as how to handle the case where the current user is not a member of the group and will we ever support multiple groups. Could you file an issue?

bouncer/views.py
```diff
@@ -144,9 +188,25 @@ def healthcheck(request):
...
 
+def _is_http_url(url):
+    return url.startswith('http://') or url.startswith('https://')
+
+
+def _pretty_url(url):
+    """
+    Return the domain name from `url` for display.
+    """
+    parsed_url = parse.urlparse(url)
```
robertknight: Good catch! Fixed.

I had a look through the `urlparse` tests and it appears that invalid IPv6 addresses are basically the only thing that `urlparse` will reject.

bouncer/views.py
robertknight: Agreed. Done.

bouncer/views.py
```diff
@@ -96,6 +92,54 @@ def index(request):
         location=request.registry.settings["hypothesis_url"])
 
 
+@view.view_config(renderer="bouncer:templates/annotation.html.jinja2",
```
robertknight: Yes, although the content in that template (visual & meta tags) specifically refers to a _single_ annotation. I think we'll probably want to revise the interstitial page a bit for this endpoint, but my plan was to solicit thoughts on that once the endpoint is usable.

OK if I defer this for now?

chdorner: Yes of course!

PR #49, opened 2017-07-04
Replace JSCS / JSHint with ESLint

robertknight:
- Replace JSCS & JSHint with ESLint and our standard config for it. Since Hound does not support custom ESLint config packages, I've just moved the command to run ESLint into `make test`.
- Fix a couple of coding style issues flagged up by ESLint (re-assigning
  a function value and use of a potentially undefined global).


seanh:
Does this mean that we're no longer using Hound at all? Both jshint and eshint are now disabled in `.hound.yml`, and Hound's Python support is disabled by default

robertknight:
> Both jshint and eshint are now disabled in .hound.yml, and Hound's Python support is disabled by default

I guess that's true. There are a bunch of flake8 violations in the Python code which it would be nice to fix. Enabling Hound for the Python code might be a good idea, even though I'd prefer to just run flake8 or other tools as part of a lint job ;)

robertknight:
This is ready for another look @seanh .

seanh:
Perfect!

PR #50, opened 2017-07-04
Update JS dev dependencies

robertknight:
Update JS dev dependencies to latest versions to fix the many
deprecation warnings during package installation.

"watchify" was added because karma-browserify now requires it as a peer
dependency.

"karma-phantomjs2-launcher" was replaced with "karma-phantomjs-launcher" because the latter now does use PhantomJS v2 and the former recommends this change during package installation.

### DIFF
--- a/karma.config.js
+++ b/karma.config.js
@@ -16,7 +16,7 @@ module.exports = function (config) {
     colors: true,
     logLevel: config.LOG_INFO,
     autoWatch: false,
-    browsers: ['PhantomJS2'],
+    browsers: ['PhantomJS'],
     singleRun: true,
     concurrency: Infinity
   });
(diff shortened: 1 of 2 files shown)


seanh:
Works here, although I'm still getting one warning from npm: `npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue`. `minimatch` isn't listed directly in `package.json`, seems to be a dependency of a dependency.

Thanks for cleaning up these warnings.

By the way should we commit the `package-lock.json` file? npm says so.

robertknight:
@seanh - Yes, `package-lock.json` should be committed, if we're using npm rather than yarn for a particular project. It is npm v5's equivalent of `yarn.lock`. Providing everyone who works on or builds the project uses npm v5, they will all end up with exactly the same versions of all dependencies. Other devs using older npm versions can still install the package but could end up with slightly differing versions that match the constraints in package.json but not the exact versions in the lockfile.

Issue #54, opened 2017-08-04
Link with deleted annotations fails


chdorner:
We don't delete annotations immediately from the search index but replace it with a `{"deleted": true}` body. We probably need to detect this at the beginning and then raise an error that should basically show the 404 error.

https://sentry.io/hypothesis/bouncer/issues/323267010/

```
KeyError: 'group'
(2 additional frame(s) were not displayed)
...
  File "pyramid/view.py", line 541, in _call_view
    response = view_callable(context, request)
  File "pyramid/config/views.py", line 353, in rendered_view
    result = view(context, request)
  File "pyramid/config/views.py", line 483, in _class_requestonly_view
    response = getattr(inst, attr)()
  File "bouncer/views.py", line 43, in annotation
    parsed_document = util.parse_document(document)
  File "bouncer/util.py", line 72, in parse_document
    group = annotation["group"]

KeyError: 'group'
```

chdorner:
@judell We missed something :) do you mind taking a look at this?

judell:
Of course, will do today.

PR #55, opened 2017-08-04
raise HTTPNotFound when annotation was deleted (issue #54)

judell:
1. Check for `{"deleted":True}` in the ElasticSearch response, and raise HTTPNotFound if so.

2. Add a test for this case.

Fixes https://github.com/hypothesis/bouncer/issues/54

### DIFF
--- a/bouncer/views.py
+++ b/bouncer/views.py
@@ -39,6 +39,9 @@ class AnnotationController(object):
             statsd.incr("views.annotation.404.annotation_not_found")
             raise httpexceptions.HTTPNotFound(_("Annotation not found"))
 
+        if ('deleted' in document['_source'] and document['_source']['deleted']):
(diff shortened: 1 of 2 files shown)


bouncer/views.py
```diff
@@ -39,6 +39,9 @@ def annotation(self):
             statsd.incr("views.annotation.404.annotation_not_found")
             raise httpexceptions.HTTPNotFound(_("Annotation not found"))
 
+        if ('deleted' in document['_source'] and document['_source']['deleted']):
```
chdorner: Might be easier and more readable to do `if document.get('deleted', False) is True`.
The current implementation, for example, would raise a 404 error when the annotation includes: `{'deleted': 'nope'}`, or any other truthy value.

seanh:
Closing in favour of https://github.com/hypothesis/bouncer/pull/57

PR #57, opened 2017-11-07
Don't crash on deleted annotations in Elasticsearch

seanh:
Fixes https://github.com/hypothesis/bouncer/issues/54

### DIFF
--- a/bouncer/util.py
+++ b/bouncer/util.py
@@ -23,6 +23,11 @@ ANNOTATION_BOILERPLATE_TEXT = (
     _('Follow this link to see the annotation in context'))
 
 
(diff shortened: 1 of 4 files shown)


robertknight:
I tested locally as follows:

1. Run bouncer service

```sh
export HYPOTHESIS_URL=http://localhost:5000
export ELASTICSEARCH_HOST=127.0.0.1
export ELASTICSEARCH_PORT=9200

make dev
```

2. Go to http://localhost:8000/<ID of existing annotation>
3. Delete the annotation from step 2
4. Go to same URL as step 2, check for an "Annotation not found" error. On master this produces a "Something went wrong" message instead.

PR #58, opened 2017-11-07
Allow the tests to run on Python 3.6

seanh:
On for example Ubuntu 17.10 Python 3.6 is installed but Python 3.5
isn't, but the tests were configured to run on 3.5 only. This changes
them to run on any Python 3.x.


robertknight:
LGTM

PR #59, opened 2017-11-08
Add a minimal flake8 config

robertknight:
This matches the config we have in the h repo.

This removes line-length related errors for those of us who have flake8 integrated with our editor ([ALE](https://github.com/w0rp/ale) FTW!)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,3 +4,6 @@ explain = true
 
 [yapf]
 based_on_style = pep8
+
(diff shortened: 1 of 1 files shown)


seanh:
I'd like to compare Python linters one day and decide which is the best to use. But in the meantime I think using the flake8 config from h everywhere makes sense and is better than nothing. Just waiting for Travis...

PR #60, opened 2017-12-04
Extract shared fixture from test cases

robertknight:
Many tests in `util_test.py` duplicated the same fixture data for fields
that are expected to be present in every ES annotation JSON document.

Extracting the fixture will make it easier to add new fields to it.

### DIFF
--- a/tests/bouncer/util_test.py
+++ b/tests/bouncer/util_test.py
@@ -53,7 +53,17 @@ def test_parse_document_returns_quote(es_annotation_doc):
     assert quote == "test_quote"
 
 
-def test_parse_document_returns_boilerplate_quote_when_no_quote(es_annotation_doc):
+@pytest.mark.parametrize('selector', [
+    # No selector (ie. a page note).
+    None,
+
+    # No quote selector. Allowed by the service even though a quote is required
+    # to anchor the annotation.
+    [{"type": "TextPositionSelector"}],
+])
+def test_parse_document_returns_boilerplate_quote_when_no_quote(es_annotation_doc, selector):
+    if selector:
+        es_annotation_doc["_source"]["target"][0]["selector"] = selector
     quote = util.parse_document(es_annotation_doc)["quote"]
     assert quote == "Hypothesis annotation for example.com"
 
--- a/tests/bouncer/util_test.py
+++ b/tests/bouncer/util_test.py
@@ -103,10 +103,11 @@ def es_annotation_doc():
     return {
         "_id": "annotation_id",
         "_source": {
-            "target": [{
-                "source": "http://example.com/example.html",
-                "selector": [{}]
-            }],
+            "target": [
+                {
+                    "source": "http://example.com/example.html",
+                }
+            ],
             "group": "__world__",
             "shared": True,
         }
--- a/tests/bouncer/util_test.py
+++ b/tests/bouncer/util_test.py
@@ -16,184 +16,96 @@ def test_parse_document_raises_if_annotated_deleted():
         })
 
 
-def test_parse_document_raises_if_no_uri():
+def test_parse_document_raises_if_no_uri(es_annotation_doc):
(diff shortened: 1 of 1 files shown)


tests/bouncer/util_test.py
```diff
@@ -16,184 +16,96 @@ def test_parse_document_raises_if_annotated_deleted():
...
+    es_annotation_doc["_source"]["target"][0]["source"] = "urn:x-pdf:the-fingerprint"
+    es_annotation_doc["_source"]["document"] = {"web_uri": "http://example.com/foo.pdf"}
+
+    document_uri = util.parse_document(es_annotation_doc)["document_uri"]
 
     assert document_uri == "http://example.com/foo.pdf"
 
 
-def test_parse_document_raises_when_uri_from_web_uri_not_string_for_pdfs():
+def test_parse_document_raises_when_uri_from_web_uri_not_string_for_pdfs(es_annotation_doc):
```
fatbusinessman: This test case seems to be testing with significantly different input than it did before; is that intentional? If so, why?

Before:

```python
{
    "_id": "annotation_id",
    "_source": {
        "target": [{
            "source": "urn:x-pdf:the-fingerprint",
            "selector": []
        }],
        "group": "__world__",
        "shared": True,
        "document": {"web_uri": 52}
    }
}
```

After:

```python
{
    "_id": "annotation_id",
    "_source": {
        "target": [{
            "source": 52,
            "selector": [{}]
        }],
        "group": "__world__",
        "shared": True,
    },
    "document": {"web_uri": "http://example.com/foo.pdf"}
}
```

tests/bouncer/util_test.py
```diff
@@ -16,184 +16,96 @@ def test_parse_document_raises_if_annotated_deleted():
...
+
+
+@pytest.fixture
+def es_annotation_doc():
+    """
+    Minimal JSON document for an annotation as returned from Elasticsearch.
+
+    This contains only fields which can be assumed to exist on all annotations.
+    """
+    return {"_id": "annotation_id",
```
fatbusinessman: I’m finding this indentation style a little hard to follow. In this case, might it be worth expanding out to a more expanded style, so all the opening and closing braces match?

tests/bouncer/util_test.py
```diff
@@ -16,184 +16,96 @@ def test_parse_document_raises_if_annotated_deleted():
...
             "_source": {
                 "target": [{
-                            "source": "urn:x-pdf:the-fingerprint",
-                            "selector": []
-                            }],
+                    "source": "http://example.com/example.html",
+                    "selector": [{}]
```
fatbusinessman: Some of the original test fixtures have this one-element list containing an empty dictionary, others have an empty list. Does that make a difference in these test cases?

tests/bouncer/util_test.py
robertknight: Gah! This was a mistake, albeit with the same outcome. The `52` should have been assigned to the "web_uri" field.

tests/bouncer/util_test.py
robertknight: I had a look through the "h" repo to try and figure out what "our" style is for nested dicts. It appears to be that the output of `jq --indent 4` is a good match, so I've changed it to that.

tests/bouncer/util_test.py
robertknight: It doesn't make a difference given the way the logic is written but it _could_, so I have 1) removed the "selector" field here since not all annotations will have a selector and 2) added a test case with a non-quote selector which is more realistic since I'm pretty sure we will have those in ES.

tests/bouncer/util_test.py
fatbusinessman: Much easier. Thank you.

fatbusinessman:
Cool. That all looks good to me; thanks for making those changes.

Either rebase your fixes into the original commits or not, as you see fit, and merge when ready.

PR #111, opened 2018-05-10
Add `make lint` to build

seanh:


seanh:
I want to add Prospector-based linting to the CI build for all our projects (including h) with a consistent Prospector config. Adding it to Bouncer as a first step, as a smaller project it'll be easier.

Marked as WIP because even in Bouncer there are many violations to fix.

PR #155, opened 2018-09-27
Avoid redirecting user to Via if target URL is known to embed client

robertknight:
If the target URL is known to embed the client, sending the user
directly to that URL will be a better experience than sending the user
through a proxy service.

There are various ways that we could detect whether a target URL embeds
the client. This commit implements the dumbest possible approach of a
hardcoded list of URL patterns.

The initial list includes a couple of Hypothesis-related URLs. Our first use for this will be in solving https://github.com/hypothesis/product-backlog/issues/814. We'll add the URL pattern(s) for that once they have been confirmed.

### DIFF
--- a/bouncer/embed_detector.py
+++ b/bouncer/embed_detector.py
@@ -1,28 +1,27 @@
+import fnmatch
 import re
 from urllib.parse import urlparse
 
-"""
-Hardcoded URL patterns where client is assumed to be embedded.
-
-Only the hostname and path are included in the pattern. The path must be
-specified.
-
-These are regular expressions so periods must be escaped.
-"""
+# Hardcoded URL patterns where client is assumed to be embedded.
+#
+# Only the hostname and path are included in the pattern. The path must be
+# specified; use "example.com/*" to match all URLs on a particular domain.
+#
+# Patterns are shell-style wildcards ('*' matches any number of chars, '?'
+# matches a single char).
 PATTERNS = [
-    "h\.readthedocs\.io/.*",
-    "web\.hypothes\.is/blog/.*",
+    "h.readthedocs.io/*",
+    "web.hypothes.is/blog/*",
 ]
 
-COMPILED_PATTERNS = [re.compile(pat) for pat in PATTERNS]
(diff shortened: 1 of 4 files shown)


bouncer/embed_detector.py
```diff
@@ -0,0 +1,45 @@
...
+
+"""
+Hardcoded URL patterns where client is assumed to be embedded.
+
+Only the hostname and path are included in the pattern. The path must be
+specified; use "example.com/*" to match all URLs on a particular domain.
+
+Patterns are shell-style wildcards ('*' matches any number of chars, '?'
+matches a single char).
+"""
```
seanh: Attribute docstrings are actually supposed to go _below_ the attribute, not above it. I'm not sure whether or not tools (e.g. Sphinx) will pick this up if it's above the attribute. To be honest, though, this isn't public API (not even in the sense of API that this module exports for other bouncer modules to use), it's only used within this module, so I think a simple code comment rather than a docstring would be fine.

bouncer/embed_detector.py
```diff
@@ -0,0 +1,45 @@
...
+PATTERNS = [
+    "h.readthedocs.io/*",
+    "web.hypothes.is/blog/*",
+]
+
+COMPILED_PATTERNS = [re.compile(fnmatch.translate(pat)) for pat in PATTERNS]
```
seanh: I'm not super familiar but I'm wondering why it's necessary to translate the fnmatch pattern into a regex? Can't fnmatch be used directly to match against the URL?

robertknight: `fnmatch` internally works by translating the shell-style wildcard into a regex which is then matched against the target string. This is a small optimization by pre-compiling the target pattern.

bouncer/embed_detector.py
robertknight: OK, I'll make it a comment.

robertknight:
>.  I think just giving people at Hypothesis an admin page to edit the list manually would be a sensible next evolution of it (if we find ourselves having to change the list often)

Agreed. For that we would need to decide where to store the list and depending on where that is, provide APIs + UIs to manage it.

PR #157, opened 2018-09-27
Fix a few lint failures and enable linting on Travis

robertknight:
Following up on https://github.com/hypothesis/bouncer/pull/154, this enables linting during the Travis build for consistency with the rest of our projects and fixes some minor code issues.


seanh:
I actually enabled stronger prospector-based linting on bouncer, and in tox, same as what the lms app has, and fixed all the lint violations: <https://github.com/hypothesis/bouncer/pull/111> I can't remember why I forgot about it - wip label was never removed and it was never merged!

PR #171, opened 2018-10-31
Upgrade to Python 3.6 and Alpine Linux 3.8

seanh:
The Alpine upgrade is needed for Python 3.6, which was added in Alpine
3.6: https://alpinelinux.org/posts/Alpine-3.6.0-released.html

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM gliderlabs/alpine:3.4
+FROM gliderlabs/alpine:3.8
 MAINTAINER Hypothes.is Project and contributors
 
 # Install system build and runtime dependencies.
-RUN apk-install ca-certificates collectd curl nodejs python3 supervisor
+RUN apk-install ca-certificates collectd curl nodejs nodejs-npm python3 supervisor
 
 # Create the bouncer user, group, home directory and package directory.
 RUN addgroup -S bouncer \
@@ -12,8 +12,7 @@ WORKDIR /var/lib/bouncer
 # Copy packaging
 COPY README.rst package.json requirements.txt ./
 
-RUN npm install --production \
-  && npm cache clean
+RUN npm install --production
 
 RUN pip3 install --no-cache-dir -U pip \
   && pip3 install --no-cache-dir -r requirements.txt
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM gliderlabs/alpine:3.4
+FROM gliderlabs/alpine:3.8
 MAINTAINER Hypothes.is Project and contributors
 
 # Install system build and runtime dependencies.
-RUN apk-install ca-certificates collectd curl nodejs python3 supervisor
+RUN apk-install ca-certificates collectd curl nodejs nodejs-npm python3 supervisor
```
seanh: Alpine's `nodejs` package no longer installs `npm`, you need `nodejs-npm` as well

Dockerfile
```diff
@@ -12,8 +12,7 @@ WORKDIR /var/lib/bouncer
 # Copy packaging
 COPY README.rst package.json requirements.txt ./
 
-RUN npm install --production \
-  && npm cache clean
+RUN npm install --production
```
seanh: Newer versions of `npm` don't like `npm cache clean`

Jenkinsfile
```diff
@@ -13,7 +13,7 @@ node {
     stage('test') {
         testApp(image: img, runArgs: '-u root') {
             sh 'pip install -q tox tox-pip-extensions'
-            sh 'cd /var/lib/bouncer && tox -e py35-tests'
+            sh 'cd /var/lib/bouncer && tox -e py36-tests'
```
seanh: Jenkins runs `tox` directly (rather than running `make`) because `make` isn't installed on Jenkins

Makefile
```diff
@@ -17,33 +17,33 @@ help:
 
 .PHONY: dev
 dev: node_modules/.uptodate
-	tox -e py35-dev
+	tox -e py36-dev
 
 .PHONY: lint
-lint:
-	tox -e py35-lint
+lint: node_modules/.uptodate
```
seanh: Make `make lint` automatically install the JavaScript dependencies if they're missing or out of date

PR #203, opened 2019-02-20
Add missing collectd-disk plugin

robertknight:
collectd was failing to start in an endless loop because the collectd disk
plugin was missing.


robertknight:
See also: https://github.com/hypothesis/h/pull/4883

It is possible this has been going on for a long time. For context, bouncer-prod had an outage for several hours this morning which may have been related: https://hypothes-is.slack.com/archives/C074BUPEG/p1550639224020000

PR #218, opened 2019-03-28
Pin Docker base image version

seanh:
To the latest version. This is to allow Dependabot's Docker support to
work.


seanh:
Right, if we pin only the minor version, then if we're on `3.8.<UNDEFINED>` and then `3.9.0` comes out then Dependabot can send us an upgrade to `3.9.0`. But if what's actually running on our servers is `3.8.1` and `3.8.2` comes out with a critical security fix, Dependabot can't send us an upgrade to that patch version because it has no way to know what patch version we currently have deployed.

PR #248, opened 2019-07-18
Silence a warning that Pyramid is triggering

seanh:
Fixes an issue that is causing the tests to fail on master currently.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,3 +1,9 @@
+[pytest]
+filterwarnings =
+    error
+    ; Suppress warnings about an import of `imp` by venusian and Pyramid
+    ignore:the imp module is deprecated in favour of importlib
+
 [tox]
 envlist = py36-tests
 skipsdist = true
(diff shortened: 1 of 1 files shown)


robertknight:
The way that the warning is suppressed here relies upon `bouncer/util.py` being the first module to be loaded that imports one of the affected packages, so I think this could break if the load order changes for some reason.

How about doing this instead:

```diff
diff --git a/tox.ini b/tox.ini
index 789a256..76896a0 100644
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,12 @@ minversion = 3.8.0
 requires = tox-pip-extensions
 tox_pip_extensions_ext_venv_update = true
 
+[pytest]
+filterwarnings =
+    error
+    ; Suppress warnings about an import of `imp` by venusian and Pyramid
+    ignore:the imp module is deprecated in favour of importlib
+
 [testenv]
 skip_install = true
 passenv =
@@ -35,7 +41,7 @@ commands =
     lint: flake8 .
     format: black bouncer tests
     checkformatting: black --check bouncer tests
-    tests: coverage run -m pytest -Werror {posargs:tests/bouncer/}
+    tests: coverage run -m pytest {posargs:tests/bouncer/}
     {docstrings,checkdocstrings}: sphinx-apidoc -ePMF -a -H "Dooccsstrinngs!!" --ext-intersphinx --ext-todo --ext-viewcode -o {envtmpdir}/rst .
     docstrings: sphinx-autobuild -BqT -z bouncer -z tests -b dirhtml {envtmpdir}/rst {envtmpdir}/dirhtml
     checkdocstrings: sphinx-build -qTn -b dirhtml {envtmpdir}/rst {envtmpdir}/dirhtml
```

seanh:
The only problem I see with the `tox.ini` approach is that it disables it system-wide, not just this one instance of it, so a new instance of this same warning wouldn't be reported. But avoiding that is probably not worth the effort, so it's fine

seanh:
Done

PR #276, opened 2019-10-03
Set default envvars in dev

seanh:
Set default values for environment variables in dev, to get bouncer working with local h and via.

Bouncer doesn't have a DB, and doesn't have any sensitive envvars that need to be set, so there's no need for bouncer to use the devdata repo. It also doesn't have a config file! So use `tox.ini` to set envvars in dev only.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,11 +16,13 @@ tox_pyenv_fallback = false
 
 [testenv]
 skip_install = true
+setenv =
+    dev: DEBUG = {env:DEBUG:yes}
+    dev: HYPOTHESIS_AUTHORITY = {env:HYPOTHESIS_AUTHORITY:localhost}
+    dev: HYPOTHESIS_URL = {env:HYPOTHESIS_URL:http://localhost:5000}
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -16,11 +16,13 @@ tox_pyenv_fallback = false
...
+setenv =
+    dev: DEBUG = {env:DEBUG:yes}
+    dev: HYPOTHESIS_AUTHORITY = {env:HYPOTHESIS_AUTHORITY:localhost}
+    dev: HYPOTHESIS_URL = {env:HYPOTHESIS_URL:http://localhost:5000}
+    dev: VIA_BASE_URL = {env:VIA_BASE_URL:http://localhost:9080}
 passenv =
     dev: CHROME_EXTENSION_ID
-    dev: DEBUG
-    dev: HYPOTHESIS_AUTHORITY
-    dev: HYPOTHESIS_URL
```
seanh: These are no longer needed in passenv since the env subsitutions in setenv now read them

seanh:
Fixed

PR #283, opened 2019-10-24
Add isort to `make format`

seanh:


tests/bouncer/search_test.py
```diff
@@ -1,8 +1,7 @@
-from mock import ANY, MagicMock, patch
-
 from elasticsearch import Elasticsearch
 
 from bouncer.search import get_client, includeme
+from mock import ANY, MagicMock, patch
```
robertknight: Looks like `mock` is sorted incorrectly here. Could this just be replaced with `unittest.mock`?

PR #314, opened 2020-03-10
Re-compile requirements.txt

seanh:
Re-run `make pip-compile` to recreate the requirements.txt file. This
doesn't change any of the requirements or their versions. It's just that
pip-tools (which `make pip-compile` uses) seems to have changed how it
"annotates" the lines in requirements.txt. So I'm just updating
requirements.txt with the new-style annotations.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,10 +4,10 @@
 #
 #    pip-compile
 #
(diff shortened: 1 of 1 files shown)


seanh:
I'm gonna merge this even though the tests are failing because the same test failure is also on master

PR #315, opened 2020-03-10
Upgrade zope.interface to its latest version

seanh:
This is the result of running:

    make upgrade-package name=zope.interface


seanh:
I think Dependabot should have been keeping this up to date for us. I don't get why it hasn't

PR #335, opened 2020-06-23
Remove Travis CI, run everything on Jenkins

seanh:


seanh:
Had a go at this but I think something is wrong with the frontend tests. Also from watching Jenkins it kind of looks like it's running things in serial not parallel, but it's hard to tell

seanh:
Bah

PR #342, opened 2020-08-03
Add GitHub Actions and remove Travis CI

seanh:


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,41 @@
...
+  frontend:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Cache the node_modules dir
+      uses: actions/cache@v2
+      with:
+        path: node_modules
+        key: ${{ runner.os }}-node_modules-${{ hashFiles('yarn.lock') }}
```
robertknight, line 37: Bouncer still uses npm rather than yarn so this should be `package-lock.json`.

seanh: Fixed

Makefile
```diff
@@ -25,8 +25,14 @@ dev: node_modules/.uptodate python
...
-lint: node_modules/.uptodate python
+lint: backend-lint frontend-lint
+
+.PHONY: backend-lint
+backend-lint: python
 	@tox -qe lint
+
+.PHONY: frontend-lint
+frontend-lint: node_modules/.uptodate
 	@./node_modules/.bin/eslint bouncer/scripts
```
robertknight, line 36: Looks like we never got around to setting up Prettier for front-end code formatting in this repo. Oh well, exercise for another time.

PR #412, opened 2021-04-16
Make the status endpoint consistent with other apps

seanh:
Every other app uses "okay" not "ok". This will allow Bouncer's Pingdom alarms to be the same as those for other apps.

It wouldn't do any harm to leave the version number in the status endpoint, but I'm not sure it has much utility and none of our other apps have it.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,4 +8,4 @@ source =
 [report]
 show_missing = True
 precision = 2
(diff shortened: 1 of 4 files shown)


robertknight (approved):
LGTM

seanh:
Rebased

PR #469, opened 2021-10-05
Replace Browserify with ES modules

robertknight:
Bouncer has only a small amount of JavaScript and used Browserify purely
as a module system provider. With our currently supported browsers and
current version of Karma we can just use native ES modules instead.
Removing Browserify usage will allow us to trim the project's
dependencies.

 - Convert `redirect.js` and `redirect-test.js` to ES modules

 - Update Karma configuration to load JS files as modules and remove
   Browserify usage

 - Remove Browserify and associated dependencies

This partially resolves the entry for Bouncer in https://github.com/hypothesis/private-issues/issues/77. Other dependencies (eg. Karma, mocha) also need to be updated to resolve the remaining issues.

----

**Testing:**

Run `make dev`, then check that if you follow an annotation share link from the client or h, the local Bouncer server (localhost:8000) correctly redirects to Via (localhost:9083) or tries to activate the extension.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,3 +1,6 @@
 {
-  "extends": "hypothesis"
+  "extends": "hypothesis",
+  "parserOptions": {
+    "sourceType": "module"
+  }
 }
--- a/bouncer/scripts/redirect.js
+++ b/bouncer/scripts/redirect.js
@@ -1,5 +1,3 @@
-'use strict';
-
 /**
  * Configuration information for the client-side code that detects the best way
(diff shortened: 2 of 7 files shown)


esanzgar (approved):
Please, could you add `browserslist` property in the `package.json`? Just to make clear which browsers we support (it wouldn't do anything else, really).

robertknight:
> Please, could you add browserslist property in the package.json? 

Will do. I'll make it the same as the new browserslist for the client.

Issue #565, opened 2022-10-25
Always deploy the correct Docker image version


seanh:
For details see https://github.com/hypothesis/via/issues/813

PR #566, opened 2022-10-27
Always deploy the correct Docker image version

seanh:
Fixes https://github.com/hypothesis/bouncer/issues/565. For details see: https://github.com/hypothesis/via/pull/814


indigobravo (approved):
👍

PR #609, opened 2023-01-19
"semanticscholar.org" => "www.semanticscholar.org"

robertknight:
I added the exact URL pattern that we were asked to add, but the reader is actually hosted at "www.semanticscholar.org" and Bouncer's matching does not currently ignore "www." as we do in some other places.


tests/unit/bouncer/embed_detector_test.py
```diff
@@ -19,6 +19,7 @@ class TestUrlEmbedsClient:
             # Example matching URLs for various sites on the list.
             "https://docdrop.org/pdf/1Vsd26C0KuMw4Mj1WEBjBz1T8G75vIhWx-gaQEE.pdf/",
             "https://docdrop.org/video/AJXGJYl0wJc/",
+            "https://www.semanticscholar.org/reader/5e331bf7887e2e634bf5b12788849d2d2b74bc7f",
```
robertknight, line 22: This is taken from an example reader URL that was shared with us in Slack.

PR #619, opened 2023-02-06
Update all instances of actions/checkout and actions/cache to v3

acelaya:
This pull request addresses a couple of warnings printed in our workflows, by updating to the latest instances of some actions and changing how we pass info between jobs via output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Setup python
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         id: list-python-versions
         run: |
           pip install hdev
-          echo "::set-output name=python_headline_version::`hdev python_version --style plain --first`"
+          echo "python_headline_version=`hdev python_version --style plain --first`" >> $GITHUB_OUTPUT
```
acelaya, line 35: See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #637, opened 2023-04-13
Use "activate" API to redirect and activate extension

robertknight:
**Depends on https://github.com/hypothesis/bouncer/pull/638**

If the installed Hypothesis Chrome extension supports the ["activate"](https://github.com/hypothesis/browser-extension/pull/1197) message,
use it to redirect the current tab and activate the extension.  This enables
bouncer to activate just one specific Hypothesis extension, if there are
multiple builds installed. The previous method would result in all installed
builds trying to activate.

In the process of using this API, the `redirect` function was converted
to be async, so that it can report when all of its async internal steps
are done. This makes writing the tests much easier. The tests have also
been rewritten to more clearly reflect the meaningfully different
scenarios that can occur.

Part of https://github.com/hypothesis/browser-extension/issues/1177

### DIFF
--- a/bouncer/scripts/redirect.js
+++ b/bouncer/scripts/redirect.js
@@ -32,6 +32,29 @@ function defaultNavigateTo(url) {
 }
 
 /**
+ * Wrapper around `chrome.runtime.sendMessage` [1] which returns a Promise.
+ *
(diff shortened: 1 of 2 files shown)


acelaya (approved):
👍🏼

PR #639, opened 2023-04-14
Support setting a different extension ID per authority

robertknight:
Allow the Chrome extension ID to be customized depending on the authority that an annotation is associated with. The extension IDs are specified by setting the `CHROME_EXTENSION_ID` env var to a JSON map of authority to extension ID. The map must contain a "default" field which specifies the extension to use if no other entry matches. The existing syntax of setting `CHROME_EXTENSION_ID` to a single ID string is still supported.

The initial use case for this is to support bouncer links for annotations created using the Hypothesis for Confluence extension. The eventual plan is to fold that integration back into the main extension, so this might be fairly short-lived flexibility.

For example:

```
export CHROME_EXTENSION_ID='{"default":"abc...", "atlassian.hypothes.is": "def..."}'
```

### DIFF
--- a/bouncer/app.py
+++ b/bouncer/app.py
@@ -1,3 +1,4 @@
+import json
 import os
 
 import pyramid.config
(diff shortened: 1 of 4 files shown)


bouncer/app.py
```diff
@@ -20,10 +21,18 @@ def settings():
...
 
+    extension_ids = os.environ.get(
+        "CHROME_EXTENSION_ID", "bjfhmglciegochdpefhhlphglcehbmek"
+    )
+    if extension_ids.strip().startswith("{"):
+        extension_ids = json.loads(extension_ids)
+        if not extension_ids.get("default"):
+            raise Exception('CHROME_EXTENSION_ID map must have a "default" key')
+    else:
+        extension_ids = {"default": extension_ids}
```
seanh, line 32: Alternatively you could do:

```python
try:
    extension_ids = json.loads(os.environ.get("CHROME_EXTENSION_ID", "..."))
except ValueError:  # (Or whatever exception json.loads raises)
    extension_ids = {"default": extension_ids}

if "default" not in extension_ids:
    raise RuntimeError("...")
```

Alternatively you could add a new envvar named `CHROME_EXTENSION_IDS` (plural) and have the code simply assume the envvar is present and is a valid JSON object and crash otherwise. Add the new envvar, then deploy the code, then remove the old envvar.

seanh (approved):

bouncer/app.py
robertknight: Thanks for the suggestions. Having a separate env var would have the advantage that the expected format is clear, though we'd need to define the precedence if both `CHROME_EXTENSION_ID` and `CHROME_EXTENSION_IDS` are set.

robertknight:
Thanks for the review. I'm going to get this merged as-is and verify that it works as expected in production.

PR #702, opened 2024-01-04
Reorganise Bouncer's Gunicorn config

seanh:
This reorganise's Bouncer's Gunicorn config files following the same pattern
as was recently done in h:

hypothesis/h#8407

I'm applying this simpler and more explicit Gunicorn config approach to
all our apps consistently. For motivation see:

https://docs.google.com/document/d/13AnUPtu9AO3PfRm-fhH7_3RZzyLGSFd4uPv36dIeynA

Benefits:

* Gunicorn config files are explicitly mentioned in the `gunicorn`
  commands in the `supervisord[-dev].conf` files. This makes things less
  confusing/surprising, less likely that a Gunicorn config file will be
  missed (compared to Gunicorn's default behavior if you run it without
  a `--config` argument: it reads any `gunicorn.conf.py` file in the
  current working directory).

* The Gunicorn config files are in the `conf/` directory with the other
  config files. Again: makes it less likely that a config file will go
  unnoticed.

* Separate production and development Gunicorn config files. This is
  consistent with our separate production and development Pyramid config
  files, and is probably a good idea to prevent development settings
  accidentally getting applied in production. (Gunicorn's default
  behavior of reading a root `gunicorn.conf.py` file would mean that
  file gets read in both dev and production).


bouncer/app.py
```diff
@@ -45,7 +45,7 @@ def settings():
     return result
 
 
-def app():
+def create_app(_=None, **_settings):
```
seanh, line 48: Renamed `app` to `create_app()` like our other apps. Bouncer was also passing Gunicorn a WSGI app rather than a WSGI app factory: the `gunicorn` command was `gunicorn bouncer.app:app()` rather than `gunicorn bouncer.app:app`. I've changed it to the factory version the same as our other apps, hence why the function needs to have arguments now although it doesn't actually pay any attention to them.

conf/development.ini
```diff
@@ -0,0 +1,30 @@
+[app:main]
```
seanh, line 1: Bouncer didn't have a `development.ini` file. I just copy-pasted the one from `test-pyramid-app`

conf/gunicorn-dev.conf.py
```diff
@@ -0,0 +1,5 @@
+bind = "0.0.0.0:8000"
+reload = True
+reload_extra_files = "bouncer/templates"
+timeout = 0
+workers = 2
```
seanh, line 5: Same exact `gunicorn-dev.conf.py` as I've used in other apps.

conf/gunicorn.conf.py
```diff
@@ -0,0 +1,2 @@
+bind = "0.0.0.0:8000"
+worker_tmp_dir = "/dev/shm"
```
seanh, line 2: Adding the `worker_tmp_dir` setting from LMS to Via, @marcospri

conf/production.ini
```diff
@@ -0,0 +1,42 @@
+[pipeline:main]
```
seanh, line 1: Bouncer didn't have a `production.ini`, I just copy-pasted the one from test-pyramid-app

conf/supervisord-dev.conf
```diff
@@ -0,0 +1,28 @@
+[supervisord]
```
seanh, line 1: Bouncer didn't have a `supervisord-dev.conf`. `tox.ini` just ran `gunicorn` directly which was possible with Bouncer because it only really has one process that needs to run in dev. I just copy-pasted the `supervisord-dev.conf` from test-pyramid-app

marcospri (approved):

seanh:
Pushed a couple of fixes to this one:

1. Added `newrelic` to Bouncer's Python requirements
2. Added `newrelic-admin run-program` to `supervisord-dev.conf` (it was already in `supervisord.conf`)

PR #705, opened 2024-01-08
Upgrade ES library

marcospri:
Found this while doing the python upgrade, much better to do it independently of that.

See commits for a little bit details on the changes.

Not sure why we never got this PR from dependabot though.


requirements/requirements.txt
```diff
@@ -1,17 +1,20 @@
...
     #   -r requirements/requirements.in
+    #   elastic-transport
     #   requests
     #   sentry-sdk
 charset-normalizer==2.0.1
     # via requests
-elasticsearch==6.3.1
+elastic-transport==8.11.0
+    # via elasticsearch
+elasticsearch==8.11.1
```
marcospri, line 17: Is there a reason for this to be in a version from 2018?

Any issues about our actual ES server version? I don't think this was pinned to one particular version so it doesn't seem intentional.

seanh: https://github.com/hypothesis/bouncer/pull/225#issuecomment-483421824

It looks like Rob was not happy with the fact the Bouncer's tests apparently mock Elasticsearch, and saw little value in updating the library anyway.

Bouncer's a pretty simple app so it shouldn't be too difficult to test it manually.

Could perhaps rewrite the tests to not mock Elasticsearch as well although that may be difficult since Bouncer talks directly to h's Elasticsearch. For example it's code in h, not Bouncer, that adds the documents to Elasticsearch that Bouncer reads.

seanh (approved):
Might want to manually test this as per Rob's old comment that I linked to

marcospri:
This is broken, at least by this error:

```TypeError: get() got an unexpected keyword argument 'doc_type'```

See https://github.com/elastic/elasticsearch-py/issues/846

We'll have to check if we can upgrade to 8.0.0 with updating the server version.

requirements/requirements.txt
robertknight: > Could perhaps rewrite the tests to not mock Elasticsearch as well although that may be difficult since Bouncer talks directly to h's Elasticsearch.

Another option is that h handles talking to ES and exposes API for bouncer to use (with suitable authentication). That's assuming that h can't just get the data from Postgres instead.

PR #715, opened 2024-01-25
Fix jinja2 import after version upgrade

marcospri:
Require 100% coverage but mark existings gaps with `nocover` like we did in H. The idea is that any changes to this codebase will require the 100% coverage an ideally the existing gaps will disappear. 

Fix jinja2 issue after the upgrade.

### DIFF
--- a/bouncer/util.py
+++ b/bouncer/util.py
@@ -1,6 +1,6 @@
 from urllib import parse
 
-import jinja2
+from markupsafe import Markup
 from pyramid import i18n
 
 _ = i18n.TranslationStringFactory(__package__)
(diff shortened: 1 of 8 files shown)


tests/unit/bouncer/util_test.py
```diff
@@ -110,6 +110,12 @@ def test_parse_document_returns_authority(es_annotation_doc):
     assert authority == "hypothes.is"
 
 
+def test_get_pretty_url_for_long_url():
```
marcospri, line 113: New test for the code with the bug

bouncer/util.py
```diff
@@ -1,6 +1,6 @@
 from urllib import parse
 
-import jinja2
+from markupsafe import Markup
```
marcospri, line 3: Bug fix, this changed after the jinaj2 upgrade.

acelaya (approved):

Issue #690, opened 2023-11-07
Always redirect to Via if annotation was made using Via's transcript annotation


dwhly:
See this slack thread for details: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1699367044627499

PR #817, opened 2024-11-18
Redirect user to Via if annotation is a transcript annotation

robertknight:
When following a bouncer link for an annotation, check if the annotation is a YouTube transcript annotation made using Via. If so, redirect the user to Via even if they have the extension installed.

Via's video player app does not currently record a specific marker indicating that it was used to create the annotation, but it is the only tool that will create an annotation with a `MediaTimeSelector` selector on annotations with document URLs of the form `https://www.youtube.com/watch?v={id}`.

Fixes https://github.com/hypothesis/product-backlog/issues/1604
Fixes https://github.com/hypothesis/bouncer/issues/690

### DIFF
--- a/bouncer/scripts/redirect.js
+++ b/bouncer/scripts/redirect.js
@@ -81,6 +81,11 @@ export async function redirect(
     return;
   }
 
+  if (settings.alwaysUseVia) {
+    navigateTo(settings.viaUrl);
+    return;
+  }
+
   const chrome = window.chrome;
   if (chrome && chrome.runtime && chrome.runtime.sendMessage) {
     // The user is using Chrome, redirect them to our Chrome extension if they
--- a/bouncer/util.py
+++ b/bouncer/util.py
@@ -92,15 +92,19 @@ def parse_document(document):
     if not text:
         text = ANNOTATION_BOILERPLATE_TEXT
 
+    has_media_time = False
+
     try:
         targets = annotation["target"]
         if targets:  # pragma: nocover
             document_uri = targets[0]["source"]
             selectors = targets[0].get("selector", [])
             for selector in selectors:
-                if selector.get("type") != "TextQuoteSelector":
-                    continue
-                quote = selector.get("exact")
+                match selector.get("type"):
+                    case "TextQuoteSelector":
+                        quote = selector.get("exact")
+                    case "MediaTimeSelector":
+                        has_media_time = True
     except KeyError:
         pass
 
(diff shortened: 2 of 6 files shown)


bouncer/views.py
```diff
@@ -90,14 +91,24 @@ def annotation(self):
...
         extension_id = settings["chrome_extension_id"].get(authority, default_extension)
 
+        # If a YouTube annotation has a media time associated, this means it
+        # was made using Via's transcript annotation tool.
+        #
+        # This means we force the use of Via, even if the extension is
+        # installed.
+        always_use_via = False
+        if document_uri.startswith("https://www.youtube.com") and has_media_time:
+            always_use_via = True
```
robertknight, line 101: We will have to update this if Via's video player gains support for other video sites in future. At that point it would be better to put a more specific marker in the annotation itself indicating which tool was used.

bouncer/scripts/redirect.js
```diff
@@ -81,6 +81,11 @@ export async function redirect(
     return;
   }
 
+  if (settings.alwaysUseVia) {
+    navigateTo(settings.viaUrl);
+    return;
+  }
```
robertknight, line 87: We could cut out the frontend in this case and just issue a 3xx redirect from the backend, but this approach makes the flow more similar across the different cases.

bouncer/util.py
```diff
@@ -133,6 +137,7 @@ def parse_document(document):
         "show_metadata": show_metadata,
         "quote": _escape_quotes(quote),
         "text": _escape_quotes(text),
+        "has_media_time": has_media_time,
     }
```
robertknight, line 141: The `parse_document` function could be changed to more modern Python at some point and return a data class instead of a dict.

marcospri (approved):

marcospri:
Looks good :+1:. 

Sorry for the delay, this was not on my radar, somehow it didn't trigger the slack review daily message.

robertknight:
> Sorry for the delay, this was not on my radar, somehow it didn't trigger the slack review daily message.

Not to worry, I hadn't pinged you because the rostering work was more important. For added context Dan asked me to look at this because the issue had come up in the context of some demos he had been giving to external parties.

Regarding the Slack daily review message, I'm not sure why this wasn't showing up. As far as I can see it doesn't meet any of the criteria for exclusion listed under https://github.com/organizations/hypothesis/settings/reminders/48861.

PR #858, opened 2025-04-11
Enable Sentry's "releases" feature

seanh:
I'm hoping this is going to get rid of "Discarded session update because
of missing release" error messages that `sentry_sdk` has started logging
in production every few seconds, see:
https://hypothes-is.slack.com/archives/C1MA4E9B9/p1744116427133249

The releases feature is also a useful feature in itself, so it'll be
good to make use of this. I think we may also need to do a little work
on our deployment scripts to enable the full Sentry releases feature
set, see:
https://docs.sentry.io/product/releases/setup/release-automation/github-actions/

### DIFF
--- a/bouncer/app.py
+++ b/bouncer/app.py
@@ -3,6 +3,8 @@ import os
 
 import pyramid.config
 
+from bouncer._version import get_version
+
 
 def settings():  # pragma: nocover
     """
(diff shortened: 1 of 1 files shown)


bouncer/app.py
```diff
@@ -53,5 +55,21 @@ def create_app(_=None, **_settings):  # pragma: nocover
...
+    #
+    # h_pyramid_sentry passes any h_pyramid_sentry.init.* Pyramid settings
+    # through to sentry_sdk.init(), see:
+    # https://github.com/hypothesis/h-pyramid-sentry?tab=readme-ov-file#settings
+    #
+    # For the full list of options that sentry_sdk.init() supports see:
+    # https://docs.sentry.io/platforms/python/configuration/options/
+    config.add_settings(
+        {
+            "h_pyramid_sentry.init.release": get_version(),
```
seanh, line 70: The other option is to just set the `SENTRY_RELEASE` envvar in Elastic Beanstalk but I don't think we have a good way to automatically set that envvar based on the release. Bouncer already has this `get_version()` function that seems to be [based on git-archive substitutions](https://github.com/hypothesis/bouncer/blob/e3e61fb340a7f636a063f32cff7758d3b572b742/bouncer/_version.py#L9-L16), so let's give that a try.

robertknight (approved):

